PR #7, opened 2017-01-24
Style cleanups

dcz-collabora:
As requested, this is the pull request containing some style cleanups in boot- and build- related functions.

Some highlights would include making code cleaner, removing strange patterns like if all(cond1, cond2), and making use of Python errors. More details are to be found in commit messages, especially 0714a56f.

### DIFF
--- a/app/utils/boot/__init__.py
+++ b/app/utils/boot/__init__.py
@@ -172,42 +172,49 @@ def save_to_disk(boot_doc, json_obj, base_path, errors):
         ERR_ADD(errors, 500, err_msg)
 
 
-def _update_boot_doc_from_json(boot_doc, json_pop_f, errors):
-    """Update a BootDocument from the provided JSON boot object.
+def _get_boot_seconds(boot_dict):
+    """Returns boot time in seconds"""
+    try:
+        boot_time_raw = boot_dict[models.BOOT_TIME_KEY]
+    except KeyError:
+        raise BootValidationError("Boot time missing")
+    try:
+        boot_time = float(boot_time_raw)
+    except ValueError:
+        raise BootValidationError("Boot time is not a number: {!r}".format(boot_time_raw))
+    if boot_time < 0.0:
+        raise BootValidationError("Found negative boot time")
+    return boot_time
+
+
+def _update_boot_doc_from_json(boot_doc, boot_dict, errors):
+    """Update a BootDocument from the provided boot dictionary.
 
     This function does not return anything, and the BootDocument passed is
     updated from the values found in the provided JSON object.
 
     :param boot_doc: The BootDocument to update.
     :type boot_doc: `models.boot.BootDocument`.
-    :param json_pop_f: The function used to pop elements out of the JSON
-    object.
-    :type json_pop_f: function
+    :param boot_dict: Boot dictionary.
+    :type boot_dict: dict
     :param errors: Where errors should be stored.
     :type errors: dict
     """
-    boot_time = json_pop_f(models.BOOT_TIME_KEY, 0.0)
+    
     try:
-        seconds = float(boot_time)
-    except ValueError, ex:
-        # Default to 0.
+        seconds = _get_boot_seconds(boot_dict)
+    except BootValidationError, ex:
         seconds = 0.0
-        err_msg = (
-            "Error reading boot time data, got: %s; defaulting to 0" %
-            str(boot_time))
+        err_msg = ("Error reading boot time data; defaulting to 0")
         utils.LOG.exception(ex)
         utils.LOG.error(err_msg)
         ERR_ADD(errors, 400, err_msg)
 
-    try:
-        if seconds < 0.0:
-            seconds = 0.0
-            ERR_ADD(errors, 400, "Boot time is negative, defaulting to 0")
-
-        if seconds == 0.0:
-            boot_doc.time = datetime.datetime(
-                1970, 1, 1, hour=0, minute=0, second=0)
-        else:
+    if seconds == 0.0:
+        boot_doc.time = datetime.datetime(
+            1970, 1, 1, hour=0, minute=0, second=0)
+    else:
+        try:
             time_d = datetime.timedelta(seconds=seconds)
             boot_doc.time = datetime.datetime(
                 1970, 1, 1,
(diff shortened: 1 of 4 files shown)


app/utils/build/__init__.py
```diff
@@ -230,106 +237,99 @@ def parse_build_data(build_data, job, kernel, errors, build_dir=None):
     :type errors: dictionary
     :param build_dir: Full path to the build directory.
     :type build_dir: string
-    :return A 2-tuple: A BuildDocument or None, a dictionary with artifacts
-    keys and names.
+    :return A 2-tuple: (BuildDocument, a "artifact type": "artifact name" dictionary)
```
MiloCasagrande: This is >80 chars.

app/utils/boot/__init__.py
```diff
@@ -33,6 +33,11 @@
 import utils.db
 import utils.errors
 
+try: # Py3K compat
```
MiloCasagrande: Either move the comment on a separate line (preferable) or add two spaces.

app/utils/boot/__init__.py
```diff
@@ -52,6 +57,8 @@
 class BootImportError(Exception):
     """General boot import exceptions class."""
 
+class BootValidationError(ValueError, BootImportError):
```
MiloCasagrande: Add one more line above here.

app/utils/boot/__init__.py
```diff
@@ -165,124 +172,125 @@ def save_to_disk(boot_doc, json_obj, base_path, errors):
...
-    :param json_pop_f: The function used to pop elements out of the JSON
-    object.
-    :type json_pop_f: function
+    :param boot_dict: Boot dictionary.
+    :type boot_dict: dict
     :param errors: Where errors should be stored.
     :type errors: dict
     """
-    boot_time = json_pop_f(models.BOOT_TIME_KEY, 0.0)
+
```
MiloCasagrande: There appears to be 4 spaces here instead of just one empty line.

app/utils/boot/__init__.py
```diff
@@ -52,6 +57,8 @@
 class BootImportError(Exception):
     """General boot import exceptions class."""
 
+class BootValidationError(ValueError, BootImportError):
+    """General error for values of boot data."""
```
MiloCasagrande: And one more line under here.

app/utils/boot/__init__.py
```diff
@@ -165,124 +172,125 @@ def save_to_disk(boot_doc, json_obj, base_path, errors):
...
-    # If the mach_alias key is defined in the JSON data, its value will be the
-    # value of the mack key.
-    mach_alias = json_pop_f(models.MACH_ALIAS_KEY, None)
-    if mach_alias:
-        boot_doc.mach = mach_alias
+    # mach_alias_key takes precedence if defined
+    boot_doc.mach = boot_dict.get(models.MACH_ALIAS_KEY,
+        boot_dict.get(models.MACH_KEY, None))
 
+def _check_for_null(board_dict):
```
MiloCasagrande: Add one more line above here.

app/utils/boot/__init__.py
```diff
@@ -165,124 +172,125 @@ def save_to_disk(boot_doc, json_obj, base_path, errors):
...
     for key in NON_NULL_KEYS:
-        t_val = str(get_func(key, ""))
-
-        val = t_val.lower()
-        if any([not val, val == "null", val == "none"]):
-            raise BootImportError(err_msg.format(key, t_val))
+        val = board_dict.get(key, None)
+        if val is None or (isinstance(val, basestring) \
+                and val.lower() in ('', 'null', 'none')):
+            raise BootValidationError("Invalid value found for mandatory key {!r}: {!r}".format(key, val))
```
MiloCasagrande: It's >80 chars.

app/utils/boot/tests/test_boot_import.py
```diff
@@ -123,58 +123,53 @@ def test_check_for_null_with_none(self):
...
     def test_check_for_null_with_none_from_string_lower(self):
         boot_report = (
             '{"job": "none", "board": "none", '
             '"kernel": "none", "defconfig": "none", "lab_name": "none"}'
         )
 
-        self.assertRaises(
-            bimport.BootImportError,
-            bimport._check_for_null, json.loads(boot_report).get)
+        self.assertRaises(bimport.BootValidationError
```
MiloCasagrande: You're missing a comma here or the tests won't even start.

app/utils/log_parser.py
```diff
@@ -530,82 +530,72 @@ def _clean_path(line):
...
         status = 500
-
+        ERR_ADD(errors, status, err_msg % (job, kernel, defconfig))
+        return status, [], [], []
+
+    def _save_lines(lines, filename):
+        # TODO: count the lines here.
+        if not lines:
+            return
+        with open(filename, mode="w") as w_file:
```
dcz-collabora: I switched from io.open() to builtin open(), as the latter is immediately recognizable to anyone familiar with Python. I didn't encounter any other project which would explicitly use io.open(), so I assumed using the clearer open() is safe here.

app/utils/build/__init__.py
```diff
@@ -230,106 +237,99 @@ def parse_build_data(build_data, job, kernel, errors, build_dir=None):
...
-        except KeyError, ex:
-            err_msg = (
-                "Missing mandatory key in build data (job: %s, kernel: %s)")
-            utils.LOG.exception(ex)
-            utils.LOG.error(err_msg, job, kernel)
-            ERR_ADD(errors, 500, err_msg % (job, kernel))
-    else:
-        utils.LOG.error("Provided data does not look like json")
-        ERR_ADD(errors, 500, "Provided data is not json")
+        build_doc.metadata = build_data
```
dcz-collabora: pop-ing works better for selecting which values were used, not values validated against. Nevertheless, a copy of the input data should be used if `pop` is the right thing to do.

app/utils/build/__init__.py
```diff
@@ -230,106 +237,99 @@ def parse_build_data(build_data, job, kernel, errors, build_dir=None):
...
-    build_doc = None
-    artifacts = {}
+    if not isinstance(build_data, types.DictionaryType):
+        raise BuildError(500, "JSON data is not a dictionary")
+
+    data = build_data
+
+    try:
+        defconfig = data[models.DEFCONFIG_KEY]
+        d_job = data.get(models.JOB_KEY, job)
```
dcz-collabora: Saving a method bound to an object reduced the code readability. It's easier to understand the line:
```
job = data.get("job")
```
compared to your example. Trying to figure out what the above line does, you only need to know what the object is - it's in the function header, and looks like a dictionary.
To know what `data_get` does, the reader must first backtrack to find where it was defined before guessing/finding out what it actually does.

The code doesn't seem performance-critical, so I replaced this with the readable versoin.

dcz-collabora:
Sorry, I didn't run the change through any linter before submitting. I didn't have the opportunity to do any serious tests - while I tried to keep the functioning exactly the same, the main intention was to help you notice what can be improved.

I have fixed issues that I did not comment on.

app/utils/build/__init__.py
MiloCasagrande: We can safely remove the pop call, it is not useful anymore and the `metadata` field is not really used anywhere.

app/utils/log_parser.py
MiloCasagrande: It's safe yes, but since `io.open` is already used in the same file, keep using it or cleanup also all the other uses of it.

app/utils/boot/tests/test_boot_import.py
```diff
@@ -81,7 +81,7 @@ def test_import_and_save_db_error(self, mock_conn):
         self.assertListEqual([500], errors.keys())
 
     def test_parse_from_json_simple(self):
-        errors = {}
+        errors = {}it's
```
MiloCasagrande: Need to remove the appended `it's`.

PR #12, opened 2017-07-24
Fix SoC name handling in LAVA v2 callback data

gctucker:
The SoC name is now passed in the LAVA v2 job definition as
'platform.mach' so read it from the callback data to populate the
'mach' meta-data for each boot.


gctucker:
This relies on https://github.com/kernelci/lava-ci-staging/pull/11 in order to have the mach name in the job data.

PR #14, opened 2017-07-28
Update the test POST API

loysollivier:
Update the test_set POST API to:
- add the created test_set as a child of test_suite

Update the test_case POST API to:
- keep test_set and test_suite references in the test_case object
- add the created test_case as a child of test_suite and test_set

### DIFF
--- a/app/handlers/test_set.py
+++ b/app/handlers/test_set.py
@@ -75,6 +75,10 @@ class TestSetHandler(htbase.TestBaseHandler):
                         response.headers = {
                             "Location": "/test/set/%s" % str(doc_id)}
 
+                        # Update test_suite add test set ID
+                        self._update_test_suite_add_test_set_id(
+                            doc_id, suite_oid, suite_name)
+
                         if test_cases:
                             if isinstance(test_cases, types.ListType):
                                 response.status_code = 202
@@ -105,6 +109,25 @@ class TestSetHandler(htbase.TestBaseHandler):
 
         return response
 
+    def _update_test_suite_add_test_set_id(self, set_oid, suite_oid, suite_name):
+        """Execute the async task to add the test set ID onto the test suite.
+
+        :param set_oid: The ID of the test set.
+        :type set_oid: bson.objectid.ObjectId
+        :param suite_oid: The ID of the test suite.
+        :type suite_oid: bson.objectid.ObjectId
+        :param suite_name: The name of the test suite.
+        :type suite_name: str
+        """
+        taskq.update_test_suite_add_test_set_id.apply_async(
+            [
+                set_oid,
+                suite_oid,
+                suite_name,
+                self.settings["dboptions"], self.settings["mailoptions"]
+            ]
+        )
+
     def _import_test_cases(self, test_cases, set_oid, suite_oid, suite_name):
         """Execute the async task to import the test cases.
 
@@ -114,6 +137,8 @@ class TestSetHandler(htbase.TestBaseHandler):
         :type set_oid: bson.objectid.ObjectId
         :param suite_oid: The ID of the test suite.
         :type suite_oid: bson.objectid.ObjectId
+        :param suite_name: The name of the test suite.
+        :type suite_name: str
         """
         taskq.import_test_cases_from_test_set.apply_async(
             [
--- a/app/handlers/test_set.py
+++ b/app/handlers/test_set.py
@@ -109,7 +109,8 @@ class TestSetHandler(htbase.TestBaseHandler):
 
         return response
 
-    def _update_test_suite_add_test_set_id(self, set_oid, suite_oid, suite_name):
+    def _update_test_suite_add_test_set_id(
+            self, set_oid, suite_oid, suite_name):
         """Execute the async task to add the test set ID onto the test suite.
 
         :param set_oid: The ID of the test set.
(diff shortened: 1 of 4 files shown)


app/handlers/test_set.py
```diff
@@ -105,6 +109,25 @@ def _post(self, *args, **kwargs):
...
+        :type suite_oid: bson.objectid.ObjectId
+        :param suite_name: The name of the test suite.
+        :type suite_name: str
+        """
+        taskq.update_test_suite_add_test_set_id.apply_async(
+            [
+                set_oid,
+                suite_oid,
+                suite_name,
+                self.settings["dboptions"], self.settings["mailoptions"]
```
MiloCasagrande: Since this is new code, and passing the database/mail options is kind of an hold and bad habit, can you please remove it from here?

You can still access the database/mail options directly from the celery app, and we avoid passing connection parameters to celery nodes that can potentially be on a different network.

app/taskqueue/tasks/test.py
```diff
@@ -207,3 +207,33 @@ def import_test_cases_from_test_set(
         set_id, suite_id, suite_name, tests_list, db_options)
     # TODO: handle errors.
     return ret_val
+
+@taskc.app.task(
+    name="update-test-suite-add-test-set-id", ignore_result=False, add_to_parent=False)
+def update_test_suite_add_test_set_id(
+        set_id, suite_id, suite_name, db_options, mail_options):
```
MiloCasagrande: You would need to remove the last two arguments from here.

app/taskqueue/tasks/test.py
```diff
@@ -207,3 +207,33 @@ def import_test_cases_from_test_set(
...
+    :param db_options: The database connection parameters.
+    :type db_options: dict
+    :param mail_options: The email system parameters.
+    :type mail_options: dict
+    :return 200 if OK, 500 in case of errors; a dictionary with errors or an
+    empty one.
+    """
+
+    ret_val, errors = tests_import.update_test_suite_add_test_set_id(
+        set_id, suite_id, suite_name, db_options, mail_options)
```
MiloCasagrande: And here, instead of `db_options` and `mail_options`, you can pass - respectively - `taskc.app.conf.db_options` and `taskc.app.conf.mail_options`.

app/utils/tests_import.py
```diff
@@ -450,38 +449,49 @@ def import_test_case(
...
                 # XXX For now, force the suite_id value.
-                json_obj[models.TEST_SUITE_ID_KEY] = suite_id
+                json_obj[models.TEST_SUITE_ID_KEY] = suite_oid
 
         try:
             test_name = j_get(models.NAME_KEY, None)
+            set_id = j_get(models.TEST_SET_ID_KEY, None)
+            if set_id:
+                json_obj[models.TEST_SET_ID_KEY] = bson.objectid.ObjectId(set_id)
+            else:
```
MiloCasagrande: I think this else clause is redundant if `TEST_SET_ID_KEY` is already `None`: `j_get` is mapped to `json_obj.get`.

app/utils/tests_import.py
```diff
@@ -450,38 +449,49 @@ def import_test_case(
...
                     "Test suite ID does not match the provided one")
                 # XXX For now, force the suite_id value.
-                json_obj[models.TEST_SUITE_ID_KEY] = suite_id
+                json_obj[models.TEST_SUITE_ID_KEY] = suite_oid
 
         try:
             test_name = j_get(models.NAME_KEY, None)
+            set_id = j_get(models.TEST_SET_ID_KEY, None)
+            if set_id:
+                json_obj[models.TEST_SET_ID_KEY] = bson.objectid.ObjectId(set_id)
```
MiloCasagrande: You can probably do:
```
set_id = bson.objectid.ObjectId(set_id)
json_obj[models.TEST_SET_ID_KEY] = set_id
```

so that later you don't have to call `bson.objectid.ObjectId(set_id)` again.

app/utils/tests_import.py
```diff
@@ -450,38 +449,49 @@ def import_test_case(
...
                     ADD_ERR(errors, 500, err_msg)
+                # If test case imported correctly reference it in the test suite
+                else:
+                    update_test_suite_add_test_case_id(
+                        doc_id, suite_oid, suite_name,
+                        db_options)
+                    #and in the test set if it exists
+                    if set_id:
+                        update_test_set_add_test_case_id(
+                            doc_id, bson.objectid.ObjectId(set_id), db_options)
```
MiloCasagrande: If you assign the `ObjectId` value to `set_id` before, you don't need to call `bson.objectid.ObjectId(set_id)` again here.

app/utils/tests_import.py
loysollivier: I thought that too, but if not specifically set to None the field is set in the database and it's value is "". Not sure it changes something though.
But I prefer to manually set it to None/null then the behavior will be the same than the other fields.

MiloCasagrande: Hmmm, weird, because soon after that code a `TestCaseDocument` object is created and if nothing is in the JSON object it will default to `None`: it means it is being pushed as an empty string.

loysollivier: Yes, looking at the previous code, it was searching for the test_set_id in a different environment (kargs). As the ID did not exist in this environment the return of k_args was None, and the object was created properly. I replicated this behavior there. 
But I agree it might need some investigation to understand why it is pushed as an empty string if the else is not there.

PR #17, opened 2017-08-09
Update lava.py

mattface:
use platform.name as the board type


MiloCasagrande:
Has it been tested on staging or is so trivial that is not needed?
I assume this should fix the problem raised on IRC.

mattface:
I havent had chance yet, i'll do a very quick test on staging this morning
Yes it fixes the problem discussed on irc last night

PR #18, opened 2017-08-11
split requirements.txt in two

montjoie:
This patch splits requirements.txt in two:
- requirements.txt for common requirements between production and dev
- requirements-dev.txt for requirements needed only by dev

Signed-off-by: Corentin LABBE <clabbe@baylibre.com>

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,0 +1,4 @@
+-r requirements.txt
+mock>0.7.2
+mongomock==1.2.0
+fakeredis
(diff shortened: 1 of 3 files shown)


requirements-dev.txt
```diff
@@ -0,0 +1,3 @@
+-r requirements.txt
+mock>0.7.2
+mongomock==1.2.0
```
MiloCasagrande: You should add `fakeredis` here as well, it is used to mock redis connections.

PR #21, opened 2017-08-23
Documentation

montjoie:


INSTALL.md
```diff
@@ -1,36 +1,26 @@
-# Ansible Local Deployment
+# Install
 
-In the repository ansible hosts file, a local host is defined and is called
-`kernel-ci-backend`. This host is provided to perform local deployment using
-ansible:
+All installation docs are now on the dedicated kernelci-backend-config
```
MiloCasagrande: You can link to the other doc here:
```
[kernelci-backend-config](https://github.com/kernelci/kernelci-backend/config/INSTALL.md)
```

Need to check if the syntax is correct, but that should do the trick. At least it's a handy way to link the two together.

montjoie: Link done. thanks

PR #23, opened 2017-09-11
T6994 fix v2 regression data

gctucker:
This cleans up a bit how errors raised by celery tasks are being handled, and runs the code to look for boot regressions when receiving a `callback/lava/` request.

### DIFF
--- a/app/handlers/callback.py
+++ b/app/handlers/callback.py
@@ -14,6 +14,7 @@
 """The RequestHandler for /callback URLs."""
 
 import tornado.web
+from celery import chain
 
 import handlers.base as hbase
 import handlers.common.query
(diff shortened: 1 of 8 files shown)


app/handlers/callback.py
```diff
@@ -14,13 +14,15 @@
...
+from celery import chain
 
 import handlers.base as hbase
 import handlers.common.query
 import handlers.common.token
 import handlers.response as hresponse
 import models
 import models.token as mtoken
+import taskqueue.tasks
 import taskqueue.tasks.callback as taskq
```
MiloCasagrande: You can remove this import, it's not used anymore.

app/utils/errors.py
```diff
@@ -11,7 +11,17 @@
...
+"""Functions and classes to handle errors data structure."""
+
+
+class BackendError(Exception):
+
+    def __init__(self, errors):
+        self.errors = errors
+
+    def __str__(self):
+        return ', '.join([' '.join([str(k)] + v)
```
MiloCasagrande: Minor nitpick: can you please use double quotes? We basically use only double quotes in the code base.

PR #24, opened 2017-09-13
T6730 trigger bisection

gctucker:
This does a bit of code clean-up and adds an initial proof-of-concept method for triggering Jenkins automated bisection jobs.

Known issues and limitations so far:
* the Jenkins parameters are fixed in the server config file
* ~it would be nicer to trigger bisection jobs in a separate task~ done
* there is no filtering to avoid bisecting false positives
* the bisection results aren't sent in any email report
* the bisection results aren't easily available from the front-end

~Note: This pull request depends on #23 to be merged first otherwise regression data won't be added to the database for v2 jobs.~ done

### DIFF
--- a/app/utils/report/boot.py
+++ b/app/utils/report/boot.py
@@ -424,8 +424,7 @@ def get_boot_data(db_options, job, branch, kernel, lab_name):
 def create_boot_report(
         job,
         branch,
-        kernel, lab_name, email_format, db_options, mail_options=None,
-        jenkins_options=None):
+        kernel, lab_name, email_format, db_options, mail_options=None):
     """Create the boot report email to be sent.
 
     If lab_name is not None, it will trigger a boot report only for that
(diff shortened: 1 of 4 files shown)


app/utils/report/boot.py
```diff
@@ -14,6 +14,7 @@
 """Create the boot email report."""
 
 import itertools
+import jenkins
```
MiloCasagrande: Hmmm... now I see.

I'm not sure I like the approach of triggering the Jenkins bisection from within the email report logic.
These two steps shouldn't be coupled together.

I think we should create a separate tasks for the Jenkins bisection that is scheduled at the same time when the email report is (and with the same timeout), after all the convoluted logic to handle duplicate email reports in the `SendHandler`.

If we need to shuffle code around to reuse what is being done in the reporting code base, that's fine, but I would prefer to keep the two separate (even if it means creating the same regression data structure twice).

gctucker: Sure, I thought I'd push this first functional version before embarking into what could turn out to be some fairly intrusive refactoring.

Presumably the changes to add extra settings for Jenkins and the added dependency on the `jenkins` module are OK?

MiloCasagrande: Yes, the Jenkins config changes are fine and even the refactoring part. We could merge those if you split the PR up into multiple ones.

gctucker:
This new versions has been rebased and takes the Jenkins options from the Celery config file.  Also it passes `pep8` and tidies up a few loose ends.

PR #28, opened 2017-09-20
Move email settings from Tornado to Celery config

gctucker:
Fix how email config options are being handled:

* remove all the mail options from the Tornado server config
* stop passing the mail options to the Celery tasks
* pass the mail options from the Celery config to the util functions

Note: a similar clean-up could be done for the `db_options`.

### DIFF
--- a/app/taskqueue/tasks/report.py
+++ b/app/taskqueue/tasks/report.py
@@ -35,7 +35,6 @@ def send_boot_report(
         email_format,
         to_addrs,
         db_options,
-        mail_options,
         cc_addrs=None, bcc_addrs=None, in_reply_to=None, subject=None):
     """Create the boot report email and send it.
 
@@ -51,8 +50,6 @@ def send_boot_report(
     :type to_addrs: list
     :param db_options: The database connection parameters.
     :type db_options: dictionary
-    :param mail_options: The options necessary to connect to the SMTP server.
-    :type mail_options: dictionary
     :param cc: The list of addresses to add in CC.
     :type cc: list
     :param bcc: The list of addresses to add in BCC.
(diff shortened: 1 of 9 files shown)


app/taskqueue/tasks/report.py
```diff
@@ -236,14 +236,16 @@ def send_multiple_emails_error(
...
 
+    mail_options = data.get("mail_options")
+
     if (txt_body or html_body) and subject:
         utils.LOG.info(
             "Sending duplicate emails report for %s-%s-%s",
             job, git_branch, kernel)
         utils.emails.send_email(
-            [taskc.app.conf.mail_options.get("error_email")],
+            [mail_options.get("error_email")],
```
MiloCasagrande: I'm not convinced this is what we really want to do, nor the correct thing to do.

The idea would be to use the `mail_options` directly taken from the celery app configuration, not passing it around (more or less in the same way we are removing all those `db_options` passing stuff).

The other thing is that when this function is called it doesn't have the `mail_options` data structure.

What you might want to do is fix the `/etc/linaro/kernelci-celery.cfg` config file to reference the `error_email` if it's not in there.

Thinking about this, it might even be useless to define the mail options in the Tornado side, since they are only really used from the Celery app (and we shouldn't pass them along).

gctucker: Well, this was merely to make the existing code actually work...

I see what you mean about moving the email options from the Tornado to the Celery config.  I can send another patch to address that.

Also, maybe the new Jenkins config should also be put into the Celery config (see #24)?  It should also only be used in tasks.

MiloCasagrande: I understand, but it would be better to fix the config file in this case since the configuration was coming from there and not from the Tornado options (even if it was passed around and it was missing that variable).

Yeah, better fix it on Celery side and provide defaults there, and probably better to do the same for #24.
Thanks!

gctucker:
Let me know if this is better and I'll resend #24 rebased on top of it, with Jenkins options in the Celery config as well.

MiloCasagrande:
This looks good, thanks! 👍

PR #29, opened 2017-10-19
Remove a LOT of the email verboseness

mattface:
DO NOT MERGE

I've just put in this pull request so we can try my branch out on staging

### DIFF
--- a/app/utils/report/templates/boot.txt
+++ b/app/utils/report/templates/boot.txt
@@ -27,54 +27,10 @@
 {%- endfor %}{# board #}
 {% endfor %}{# defconfig #}
 {%- endfor %}{# arch #}
+{% else %}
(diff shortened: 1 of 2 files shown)


MiloCasagrande:
I manually applied this on staging.

gctucker:
Do we still want to shorten the build and boot reports?

khilman:
IMO, yes.  I think email reports should focus on red flags, like new regressions and failures, and summarize the rest, with links pointing to the frontend.

mattface:
Yes I believe we do, though this review should be abandoned and reattempted properly

gctucker:
@mattface How about closing this now as it's way behind, and maybe opening a new one in the light of recent discussions around verbosity of build reports?

PR #32, opened 2017-10-27
Allow OPTIONS requests on the server

loysollivier:
Allow the OPTIONS request to the server.
Send back the Origin for crossdomain security.
This option is needed for preflight requests from web browser sandboxes.

Signed-off-by: lollivier <lollivier@baylibre.com>


MiloCasagrande:
Changes look good, I'm wondering how this will work through a reverse proxy though. nginx blocks by default all CORS requests for example, and kernelci.org is not configured to accept those.

For your use case, are you behind a reverse proxy?

loysollivier:
I am not sure what to answer. I used ansible to deploy the backend and frontend onto our server. We did not do any special tweaking after that. 

What I can tell you is that it works right away with the ansible default configuration.

MiloCasagrande:
Well, changes look good and they don't do any harm. :-)
Going to merge.

PR #33, opened 2017-11-06
T7434 test plan in boot docs

gctucker:
* add a field with the test plan name in new boot document entries in the database.
* clean up (syntax issues and docstrings)

### DIFF
--- a/app/utils/callback/lava.py
+++ b/app/utils/callback/lava.py
@@ -78,7 +78,7 @@ BL_META_MAP = {
 }
 
 
-def _get_lava_test_case_data(meta, tc_data, job_data, META_DATA_MAP):
+def _get_lava_test_case_data(meta, tc_data, job_data, meta_data_map):
     """Parse the lava test definition data
 
     Parse the data from the LAVA v2 job definition sent with the callback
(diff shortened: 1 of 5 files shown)


app/utils/callback/lava.py
```diff
@@ -68,6 +68,7 @@
     models.FASTBOOT_KEY: "platform.fastboot",
     models.INITRD_KEY: "job.initrd_url",
     models.BOARD_KEY: "platform.name",
+    models.PLAN_KEY: "test.plan",
```
loysollivier: So this is a new metadata field passed by LAVA ? Could you elaborate on what this test.plan is ? Is it just a metadata field that should be in the LAVA job description or is it "created" by LAVA one doing some testing ?

gctucker: This is already passed by LAVA in the job meta-data, it is in fact set when the LAVA job is submitted.  Typically, a kernelci.org job will have this set to "boot".  As we're adding more tests, other test plans such as "kselftest" are being used.

loysollivier: OK thanks for the info

gctucker:
Update: this version will create separate boot entries when different test plans are being used.  This is fine as long as builds aren't used with more than one test plan, otherwise several boot entries will be created and the front-end is not designed to show this information, resulting in duplicate boot data.

One way to resolve this is to consolidate "boots" and "tests" to simply treat "boot" as a test plan and stop making the distinction between the two.

gctucker:
After some more discussion, we've decided to stop creating a boot entry for each test run and instead store the boot data directly into the test entry.  So this PR can be dropped.

I've created #35 with just the clean-up patch as that won't hurt.

PR #36, opened 2017-11-20
T7565 device type in bisection

gctucker:
This is to fix the issue whereby some Jenkins bisection jobs get triggered with the wrong target name.  The code in lava-ci defines `platform.name` to be either the device tree name or the device type name in an unpredictable way.  So this field can't be used to run the tests again on the same target.  The `device.type` field is consistent but it needs to be added to the database entries and sent to Jenkins.

Also clean a few things up in the Jenkins bisection job parameters.


gctucker:
Rebased, tests and doc updated.

doc/schema/1.1/get_boot.json
```diff
@@ -34,6 +34,10 @@
             "type": "string",
             "description": "The instance identifier for the board"
         },
+        "device_type": {
+            "type": "string",
+            "description": "The device type associaged with the board"
```
MiloCasagrande: Minor typo. :-)

s/associaged/associated

gctucker:
Thanks for the review.  Typo fixed, coffee brewing.

PR #40, opened 2017-11-20
T7495 Look for early LAVA v2 job failures and set them as UNKNOWN

gctucker:
In some cases, an early step in the LAVA v2 jobs may fail even before
trying to boot the kernel.  These should result in an UNKNOWN test
result rather than FAIL.  As they are not all reported as
infrastructure error by LAVA, look for known error messages in the
meta-data.

### DIFF
--- a/app/utils/callback/lava.py
+++ b/app/utils/callback/lava.py
@@ -14,6 +14,7 @@
 import errno
 import models
 import os
+import re
 import yaml
 
 import utils
@@ -81,6 +82,15 @@ META_DATA_MAP_BOOT = {
     models.DEVICE_TYPE_KEY: "device.type",
 }
 
+# This is used to detect that an early job step timed out and set the status to
+# UNKNOWN.  It should eventually be reported as an infrastructure error by
+# LAVA, at which point this code could be removed.
+TIMEOUT_RE = re.compile(r"^(.*) timed out after [\d]* seconds$")
+EARLY_TIMEOUT_STEPS = [
+    "http-download",
+    "bootloader-commands",
+]
+
 BL_META_MAP = {
     "ramdisk_addr": "initrd_addr",
     "kernel_addr": "loadaddr",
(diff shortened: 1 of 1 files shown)


app/utils/callback/lava.py
```diff
@@ -70,6 +71,15 @@
     models.BOARD_KEY: "platform.name",
 }
 
+# This is used to detect that an early job step timed out and set the status to
+# UNKNOWN.  It should eventually be reported as an infrastructure error by
+# LAVA, at which point this code could be removed.
+TIMEOUT_RE = re.compile(r"^(.*) timed out after [\d]* seconds$")
+EARLY_TIMEOUT_STEPS = [
+    "http-download",
+    "bootloader-commands",
```
mattface: The problem is with kernels that fail between the end of the bootloader-commands, and the start of auto-login-action. The ones that (in uboot as an example) output nothing after "Starting kernel". LAVA will report the bootloader-commands step as the failure as it hasn't matched the line which ends that action.

gctucker: OK I need to look at more failed logs.

When the bootloader prompts are set correctly and the bootloader does get stuck, a bootloader-commands timeout should still be treated as UNKNOWN.  I'll see if I can get some other reliable meta-data in the case where this is actually a kernel error.

mattface: Yes, your problem is that LAVA doesn't differentiate between some of the commands going wrong, and the kernel just not booting. I think it's a failing in LAVA, I'll bring it up at the team meeting tomorrow, we should be able to give a result for this.

gctucker: Adding this to the default set of bootloader prompts would already help a lot I guess (copying from some device dictionary here):
```
{% set boot_message = 'Starting kernel' %}
```

mattface: Yes for uboot devices adding that is probably all that's required.
GRUB nor iPXE output anything, so my thinking was to log a result once LAVA has sent all of the boot commands and if you've got there successfully, you have to assume anything after that is a kernel error. I'll see what other ideas the team have tomorrow.

MiloCasagrande:
I hold off merging this since I'm missing a piece of information on the LAVA side.

gctucker:
Related discussion: https://lists.linaro.org/pipermail/lava-users/2017-November/000599.html

MiloCasagrande:
OK, I guess this will have to wait until the LAVA changes are in place.

khilman:
Accoding to @mattface , LAVA changes have been merged.

gctucker:
Great, I'll get the bisection PRs out of the way first then rebase and test this against my local LAVA instance. @mattface @khilman

mattface:
This is no longer needed for latest LAVA. A timeout in any of the "pre kernel" steps will now cause an Infrastructure error, which the backend is already handling.

gctucker:
Closing this PR as the LAVA changes make this unnecessary.

PR #41, opened 2017-11-21
T7830 Fix boot regressions issue due to board instance name

gctucker:
This may be a matter of opinion or consistency, but in practice using the board instance name effectively hides away some regressions.  I see this easily with my 2 qemu devices, they run jobs alternatively so if I start with a "good" boot run on qemu0 and then run a "bad" boot on qemu1, it won't be detected as a regression unless this change is applied (or I'm lucky and an earlier good boot of the same kind was run on qemu1).

### DIFF
--- a/app/utils/boot/regressions.py
+++ b/app/utils/boot/regressions.py
@@ -317,7 +317,6 @@ def check_and_track(boot_doc, db_options):
     # it's a new regression that we need to track.
     spec = {
         models.ARCHITECTURE_KEY: b_get(models.ARCHITECTURE_KEY),
-        models.BOARD_INSTANCE_KEY: b_get(models.BOARD_INSTANCE_KEY),
         models.BOARD_KEY: b_get(models.BOARD_KEY),
         models.COMPILER_VERSION_EXT_KEY:
             b_get(models.COMPILER_VERSION_EXT_KEY),


MiloCasagrande:
It probably isn't just that simple change, but it needs to be better tested even with old regressions data.

You should check as well the `check_prev_regression`, `gen_regression_keys`, `get_regressions_by_key` functions because they all assume a `board_instance` especially for checking if we are already tracking the regression, or we might end up sort of resetting the regressions tracking. The key to track the regressions is made with all those "unique" fields, `board_instance` included.

The other option would be to check for the type of the board and if it's qemu (or similar) just don't use the `board_instance` field (where it makes sense not to use it). But that would need to be carried around everywhere, and might end up resetting the regression tracking for what we have now.

gctucker:
OK I'll take a look.

I mentioned qemu in my comment because that's what I typically use to initially test backend changes, but it applies to any board type.

Also while we need to be sure we don't lose previous regression data, I believe using board instances rather than types actually made us miss quite a few regressions.

MiloCasagrande:
That approach I think will complicate things a little bit more: right now we just look for 1 previous boot report (via `find_one3`), but if you have more than one board instance, you just get the first result based on the search criteria (and likely based on the creation date). That might give you a regression or not, depending on which boot reports came in first.

You'll likely have to look for all previous boot reports, group them based on their creation date, pick the closest group by creation date, and then decide how to consider a regression: if the one you are starting from is a FAIL, and you have >1 instances of the same board, should all of the previous ones be PASS for it to be a real regression? What is the percentage of previous PASS reports for it to be considered a regression? And if one of the previous ones was an infrastructure FAIL or just a glitch with that board? (probably we want to introduce more statuses to distinguish between the fails and categorize them correctly)

gctucker:
I'm not following this, we're still using the `models.BOARD_KEY: b_get(models.BOARD_KEY)` field to only get boots of a given board type.  So it should just return what we want, i.e. the last boot with the same lab / config / tree (job) / board type.

So I don't think we need to do any fuzzy matching with a range of previous boots, unless I'm missing something?

MiloCasagrande:
That is true if you have a single instance of a board in your lab. If you have more boards of the same model it's not true anymore, that's why we look also for the `board_instance`.

If you look for a `juno` board, but you have 4 of it (with different `board_instance` values), how do you determine if the previous boot reports are all good based only on one? Do you just base you search on a single result? You should be looking at all the previous ones, and decided how to proceed from there.

gctucker:
But there is only one last boot entry with a given combination of job / defconfig / lab / branch so the last one received according to the time stamp is the one we need.  Each job is run only once for a given board type, LAVA picks one board instance randomly (or generally speaking, the scheduler will pick one arbitrarily).  I can put together some example with real data to help explain this if you want.

MiloCasagrande:
IIRC the `board_instance` field is one of the unique fields used to assert if a boot has already been pushed, I think that's why we introduced it.

So, if that field is used to distinguish between boots, there might be multiple ones for the same board; if it's not used no, then there will always be a single one (though now I don't understand where that is used, apart from a metadata POV).

gctucker:
@MiloCasagrande Could we please try this on staging?  It would be good to check whether previous regressions get reset or if there are any undesirable side effects.

MiloCasagrande:
I prefer not to commit this directly into master if we want to test it out.

I created a staging branch, and will try to merge it in there (or if you can try to rebase on the staging branch). If I edit this PR and change the branch it tells me some commits might be lost: not sure if it's just a warning and nothing should happen or what.

I'll try to do it tomorrow from the command line in case.

gctucker:
Thanks, I'll run some regular tests on staging to see if existing regressions are still tracked.

PR #44, opened 2017-11-28
T7885 Separate tests from boots (consolidation)

gctucker:
This is a first series of changes for review before making more code changes.  It's essentially to clean up the code and make it easier to maintain in preparation for the refactoring of the `callback/lava/test` handler.  The aim being to separate test entries from boot entries, rather than have tests depend on boots.

### DIFF
--- a/app/utils/callback/lava.py
+++ b/app/utils/callback/lava.py
@@ -37,9 +37,17 @@ LAVA_JOB_RESULT = {
     CANCELING: "UNKNOWN",
 }
 
-DATA_MAP_TEST_CASE = {
+TEST_CASE_MAP = {
     models.NAME_KEY: "name",
     models.STATUS_KEY: "result",
+}
+
+TEST_CASE_NAME_EXTRA = {
+    "http-download": ["label"],
+    "git-repo-action": ["commit", "path"],
+    "test-overlay": ["name"],
+    "test-runscript-overlay": ["name"],
+    "test-install-overlay": ["name"],
 }
 
 META_DATA_MAP_TEST_SUITE = {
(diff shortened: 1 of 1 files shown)


app/utils/callback/lava.py
```diff
@@ -320,16 +211,17 @@ def _add_boot_log(meta, log, base_path):
 
     dir_path = os.path.join(
         base_path,
-        meta["job"],
-        meta["git_branch"],
-        meta["kernel"],
-        meta["arch"],
-        meta["defconfig_full"],
-        meta["lab_name"])
+        meta[models.JOB_KEY],
```
loysollivier: nitpick
I was under the impression that data extracted from the callback json was accessed using "dict string" and data from the kCI database was accessed models.STRING.

I thought it would be clearer to know if you're accessing the dB or the callback data that way. 

But overall I'd agree on using "models.STRING" all the time if it is a 1:1 matching.

app/utils/callback/lava.py
```diff
@@ -400,6 +292,58 @@ def add_boot(job_data, lab_name, db_options, base_path=utils.BASE_PATH):
...
+def _parse_lava_test_data(test_case, test_meta):
+    """Parse the test job meta-data for the lava step
+
+    :param test_case: The parsed test case data to be stored in the database.
+    :type test_case: dict
+    :param test_meta: The test meta-data from the callback.
+    :type test_meta: dict
+    """
+    test_case_name = test_case[models.NAME_KEY]
+    extra = TEST_CASE_NAME_EXTRA.get(test_case_name)
```
loysollivier: Nice one

app/utils/callback/lava.py
gctucker: The `meta` dictionary is not directly the JSON data received from LAVA but a dictionary created by parsing the JSON data, so these keys are not used when accessing the incoming data; the `META_DATA_MAP_*`s are used to make this key translation.  Also the code that creates the DB entries uses the same keys from the models, as in `app.utils.boot._parse_boot_from_json`.  

So I think it's fine, unless I'm missing something - in which case I agree, using the model keys in the incoming JSON data would be a bit wrong and work mostly out of luck.

gctucker:
@loysollivier Thanks for the review.  About testing this live, I've run it with some stub test plans using qemu on my VM but I guess we could do more real-world testing on staging.

@MiloCasagrande Please let us know if this works for you and if/when this can be deployed on staging.

MiloCasagrande:
@gctucker I'm going to review this tomorrow.
I'm going to tag a new release, deploy in production and then pick up the rest of the reviews.

gctucker:
@MiloCasagrande Sounds good, thanks.

Once this has been merged, I'll send another Github PR with some functional changes to add the boot data to the test entries.  See my work-in-progress branch here:
https://gitlab.collabora.com/gtucker/kernelci-backend/commits/T7885-separate-tests-from-boots

PR #45, opened 2017-12-11
T7885 Separate tests from boots

gctucker:
Separate test entries in the database from boot entries so they can evolve separately and avoid any cross-talk between test and boot results.

* add fields to the test entries with some boot meta-data
* always create a test entry when receiving either a test or a boot callback
* do not create a boot entry when receiving a test callback
* use the test suite name in the generated logs
* only look for regressions when receiving a boot callback for now

### DIFF
--- a/app/handlers/callback.py
+++ b/app/handlers/callback.py
@@ -143,26 +143,28 @@ class LavaCallbackHandler(CallbackHandler):
         return models.LAVA_CALLBACK_VALID_KEYS.get(method, None)
 
     def _execute_callback(self, lab_name, **kwargs):
-        response = hresponse.HandlerResponse()
         action = kwargs["action"]
         json_obj = kwargs["json_obj"]
-        tasks = []
+        response = hresponse.HandlerResponse()
+        response.status_code = 202
+        response.reason = "Request accepted and being processed"
 
         if action in ["boot", "test"]:
-            tasks.append(taskqueue.tasks.callback.lava_boot.s(
-                json_obj, lab_name))
-            tasks.append(taskqueue.tasks.boot.find_regression.s())
-
-        if action == "test":
-            tasks.append(taskqueue.tasks.callback.lava_test.s(
-                json_obj, lab_name))
-
-        if not tasks:
+            tasks = [
+                taskqueue.tasks.callback.lava_test.s(json_obj, lab_name),
+            ]
+            chain(tasks).apply_async(
+                link_error=taskqueue.tasks.error_handler.s())
+        else:
             response.status_code = 404
             response.reason = "Unsupported LAVA action: {}".format(action)
-        else:
-            response.status_code = 202
-            response.reason = "Request accepted and being processed"
+
+        # Also run the legacy boot callback to generate boot entries
+        if action == "boot":
+            tasks = [
+                taskqueue.tasks.callback.lava_boot.s(json_obj, lab_name),
+                taskqueue.tasks.boot.find_regression.s(),
+            ]
             chain(tasks).apply_async(
                 link_error=taskqueue.tasks.error_handler.s())
 
--- a/app/models/test_suite.py
+++ b/app/models/test_suite.py
@@ -52,7 +52,6 @@ class TestSuiteDocument(modb.BaseDocument):
         self.arch = None
         self.board = None
         self.board_instance = None
-        self.boot_id = None
         self.defconfig = None
         self.defconfig_full = None
         self.definition_uri = None
@@ -119,7 +118,6 @@ class TestSuiteDocument(modb.BaseDocument):
             models.ARCHITECTURE_KEY: self.arch,
             models.BOARD_INSTANCE_KEY: self.board_instance,
             models.BOARD_KEY: self.board,
-            models.BOOT_ID_KEY: self.boot_id,
             models.BUILD_ID_KEY: self.build_id,
             models.CREATED_KEY: self.created_on,
             models.DEFCONFIG_FULL_KEY: self.defconfig_full or self.defconfig,
--- a/app/models/test_suite.py
+++ b/app/models/test_suite.py
@@ -28,42 +28,61 @@ class TestSuiteDocument(modb.BaseDocument):
     A test suite is a document that can store test cases and/or test sets ran.
     """
 
-    def __init__(self, name, lab_name, build_id=None, version="1.0"):
+    def __init__(self, name, lab_name):
         """The test suite document.
 
         :param name: The name given to this test suite.
         :type name: string
         :param lab_name: The name of the lab running this test suite.
         :type lab_name: str
-        :param build_id: The ID of the defconfig associated with this
-        test suite.
-        :type build_id: str
-        :param version: The version of the JSON schema of this test suite.
-        :type version: str
         """
-        self._created_on = None
-        self._id = None
         self._name = name
-        self._version = version
-
-        self.build_id = build_id
-        self.lab_name = lab_name
+        self._lab_name = lab_name
+        self._build_id = None
+        self._version = None
+        self._id = None
+        self._created_on = None
 
         self.arch = None
         self.board = None
         self.board_instance = None
+        self.boot_log = None
+        self.boot_log_html = None
+        self.boot_result_description = None
+        self.compiler = None
+        self.compiler_version = None
+        self.compiler_version_full = None
+        self.cross_compile = None
         self.defconfig = None
         self.defconfig_full = None
         self.definition_uri = None
+        self.device_type = None
+        self.dtb = None
+        self.dtb_addr = None
+        self.endian = None
+        self.file_server_resource = None
+        self.file_server_url = None
         self.git_branch = None
+        self.git_commit = None
+        self.git_describe = None
+        self.git_url = None
+        self.initrd_addr = None
         self.job = None
         self.job_id = None
         self.kernel = None
+        self.kernel_image = None
+        self.kernel_image_size = None
+        self.load_addr = None
+        self.mach = None
         self.metadata = {}
+        self.qemu = None
+        self.qemu_command = None
+        self.retries = 0
         self.test_case = []
         self.test_set = []
         self.time = -1
         self.vcs_commit = None
+        self.warnings = 0
 
     @property
     def collection(self):
@@ -74,24 +93,22 @@ class TestSuiteDocument(modb.BaseDocument):
         """The name of the test suite."""
         return self._name
 
-    @name.setter
-    def name(self, value):
-        """Set the name of the test suite.
-
-        :param value: The name of the test suite.
-        :type value: string
-        """
-        self._name = value
+    @property
+    def lab_name(self):
+        """The name of the test lab."""
+        return self._lab_name
 
     @property
-    def id(self):
-        """The ID of the test suite as registered in the database."""
-        return self._id
+    def build_id(self):
+        """The ID of the build."""
+        return self._build_id
 
-    @id.setter
-    def id(self, value):
-        """Set the test suite ID."""
-        self._id = value
+    @build_id.setter
+    def build_id(self, value):
+        """Set the ID of the build."""
+        if self._build_id:
+            raise AttributeError("Build ID already set")
+        self._build_id = value
 
     @property
     def version(self):
@@ -101,7 +118,21 @@ class TestSuiteDocument(modb.BaseDocument):
     @version.setter
     def version(self, value):
         """Set the schema version of this test suite."""
+        if self._version:
+            raise AttributeError("Schema version already set")
         self._version = value
+
+    @property
+    def id(self):
+        """The ID of the test suite as registered in the database."""
+        return self._id
+
+    @id.setter
+    def id(self, value):
+        """Set the test suite ID."""
+        if self._id:
+            raise AttributeError("ID already set")
+        self._id = value
 
     @property
     def created_on(self):
@@ -111,36 +142,59 @@ class TestSuiteDocument(modb.BaseDocument):
     @created_on.setter
     def created_on(self, value):
         """Set the creation date of this test suite."""
+        if self._created_on:
+            raise AttributeError("Creation date already set")
         self._created_on = value
 
     def to_dict(self):
-        test_suite = {
+        return {
             models.ARCHITECTURE_KEY: self.arch,
             models.BOARD_INSTANCE_KEY: self.board_instance,
             models.BOARD_KEY: self.board,
+            models.BOOT_LOG_KEY: self.boot_log,
+            models.BOOT_LOG_HTML_KEY: self.boot_log_html,
+            models.BOOT_RESULT_DESC_KEY: self.boot_result_description,
             models.BUILD_ID_KEY: self.build_id,
+            models.COMPILER_KEY: self.compiler,
+            models.COMPILER_VERSION_FULL_KEY: self.compiler_version_full,
+            models.COMPILER_VERSION_KEY: self.compiler_version,
+            models.CROSS_COMPILE_KEY: self.cross_compile,
             models.CREATED_KEY: self.created_on,
             models.DEFCONFIG_FULL_KEY: self.defconfig_full or self.defconfig,
             models.DEFCONFIG_KEY: self.defconfig,
             models.DEFINITION_URI_KEY: self.definition_uri,
+            models.DEVICE_TYPE_KEY: self.device_type,
+            models.DTB_KEY: self.dtb,
+            models.DTB_ADDR_KEY: self.dtb_addr,
+            models.ENDIANNESS_KEY: self.endian,
+            models.FILE_SERVER_RESOURCE_KEY: self.file_server_resource,
+            models.FILE_SERVER_URL_KEY: self.file_server_url,
             models.GIT_BRANCH_KEY: self.git_branch,
+            models.GIT_COMMIT_KEY: self.git_commit,
+            models.GIT_DESCRIBE_KEY: self.git_describe,
+            models.GIT_URL_KEY: self.git_url,
+            models.ID_KEY: self.id,
+            models.INITRD_ADDR_KEY: self.initrd_addr,
             models.JOB_ID_KEY: self.job_id,
             models.JOB_KEY: self.job,
             models.KERNEL_KEY: self.kernel,
+            models.KERNEL_IMAGE_KEY: self.kernel_image,
+            models.KERNEL_IMAGE_SIZE_KEY: self.kernel_image_size,
             models.LAB_NAME_KEY: self.lab_name,
+            models.LOAD_ADDR_KEY: self.load_addr,
+            models.MACH_KEY: self.mach,
             models.METADATA_KEY: self.metadata,
+            models.QEMU_KEY: self.qemu,
+            models.QEMU_COMMAND_KEY: self.qemu_command,
+            models.RETRIES_KEY: self.retries,
             models.NAME_KEY: self.name,
             models.TEST_CASE_KEY: self.test_case,
             models.TEST_SET_KEY: self.test_set,
             models.TIME_KEY: self.time,
             models.VCS_COMMIT_KEY: self.vcs_commit,
-            models.VERSION_KEY: self.version
+            models.VERSION_KEY: self.version,
+            models.WARNINGS_KEY: self.warnings,
         }
-
-        if self.id:
-            test_suite[models.ID_KEY] = self.id
-
-        return test_suite
 
     @staticmethod
     def from_json(json_obj):
@@ -149,16 +203,10 @@ class TestSuiteDocument(modb.BaseDocument):
             local_obj = copy.deepcopy(json_obj)
             doc_pop = local_obj.pop
 
-            suite_id = doc_pop(models.ID_KEY, None)
-
             try:
                 name = doc_pop(models.NAME_KEY)
                 lab_name = doc_pop(models.LAB_NAME_KEY)
-                build_id = doc_pop(models.BUILD_ID_KEY)
-
-                test_suite = TestSuiteDocument(name, lab_name, build_id)
-                test_suite.id = suite_id
-
+                test_suite = TestSuiteDocument(name, lab_name)
                 for key, val in local_obj.iteritems():
                     setattr(test_suite, key, val)
             except KeyError:
--- a/app/utils/callback/lava.py
+++ b/app/utils/callback/lava.py
@@ -344,8 +344,7 @@ def _get_test_case_data(suite_results, suite_name):
     return test_cases
 
 
-def add_tests(job_data, lab_name, boot_doc_id,
-              db_options, base_path=utils.BASE_PATH):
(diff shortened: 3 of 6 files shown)


app/models/test_suite.py
```diff
@@ -112,55 +142,71 @@ def created_on(self):
...
             models.DEFCONFIG_KEY: self.defconfig,
             models.DEFINITION_URI_KEY: self.definition_uri,
+            models.DEVICE_TYPE_KEY: self.device_type,
+            models.DTB_KEY: self.dtb,
+            models.DTB_ADDR_KEY: self.dtb_addr,
+            models.ENDIANNESS_KEY: self.endian,
+            models.FILE_SERVER_RESOURCE_KEY: self.file_server_resource,
+            models.FILE_SERVER_URL_KEY: self.file_server_url,
             models.GIT_BRANCH_KEY: self.git_branch,
+            models.GIT_COMMIT_KEY: self.git_commit,
```
loysollivier: Shouldn't we align on this GIT_COMMIT_KEY vs VCS_COMMIT_KEY (line 194) ? Seems like a duplicate

app/utils/callback/lava.py
```diff
@@ -383,15 +402,17 @@ def add_tests(job_data, lab_name, boot_doc_id,
...
             models.NAME_KEY: suite_name,
         }
 
         utils.LOG.info("Processing test suite: {}".format(suite_name))
 
         try:
+            _get_job_meta(suite_data, job_data)
             _get_definition_meta(suite_data, job_data, META_DATA_MAP_TEST_SUITE)
+            _get_lava_meta(suite_data, job_data)
+            _add_boot_log(suite_data, job_data["log"], base_path, suite_name)
```
loysollivier: I think the frontend will need to be updated to search for this log file.
Today, it searches for the boot_id and gets the log from the boot document.

gctucker: The boot log will still be generated as before, and available in the "Boots" section on the front-end.

The extra boot fields from the test entries as well as the boot log generated by the test handler are not yet used by the front-end, but this is the next step: improving test results visualisation.

It's necessary to create the logs twice at the moment to make the tests completely separate from boots.  The idea would be to stop phase out and stop using boot entries but only use tests at some point.

app/models/test_suite.py
gctucker: Yes, it is a duplicate.  I left both to be compatible with current boot entries and also preserve the fields currently used in test entries.

Do we really need `VCS_COMMIT_KEY`?

app/handlers/callback.py
```diff
@@ -143,26 +143,28 @@ def _valid_keys(method):
...
+                link_error=taskqueue.tasks.error_handler.s())
+        else:
             response.status_code = 404
             response.reason = "Unsupported LAVA action: {}".format(action)
-        else:
-            response.status_code = 202
-            response.reason = "Request accepted and being processed"
+
+        # Also run the legacy boot callback to generate boot entries
+        if action == "boot":
```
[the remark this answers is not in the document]
gctucker: Yes.

app/models/test_suite.py
loysollivier: I don't think so, I've never seen it used, or even set...

gctucker: OK then I think we can just drop `VCS_COMMIT_KEY` from the test entries.  I'll add another patch for that.

app/utils/callback/lava.py
loysollivier: OK, I understand your point. All I am saying is that the frontend is searching for "boot_id" today in test entries. From there it looks for the boot log. 

boot_id being removed the frontend will need to be updated to use this new log. Which is actually better than reusing the same log.

gctucker: Ah sorry, you meant in the Tests views - right.  As long as both the front-end and back-end are deployed in sync it should be fine I guess.  @MiloCasagrande, please let us know if you think there are any issues with doing this.

@loysollivier Are you OK with making the corresponding front-end changes?

loysollivier: Yes, I will try to send a PR corresponding for the frontend tomorrow

app/models/test_suite.py
```diff
@@ -28,43 +28,61 @@ class TestSuiteDocument(modb.BaseDocument):
...
+        self._version = None
+        self._id = None
+        self._created_on = None
 
         self.arch = None
         self.board = None
         self.board_instance = None
-        self.boot_id = None
+        self.boot_log = None
+        self.boot_log_html = None
```
loysollivier: This looks like a lot of duplicates, why not creating a parent document such as "test_environment". That would hold all the info relative to the board / testing environment. After all, all these fields would be identical for a given serie of tests, that would be a lot of wasted space in the DB.

The backend could create one element of test_environment for each callback to test. Then all the test suites would be childs of this parent element.
 
I'm thinking about these fields specifically:
- _lab_name
- _build_id
- arch
- board
- boot_log
- boot_log_html
- boot_result_description
- compiler
- compiler_version
- compiler_version_full
- cross_compile
- defconfig
- defconfig_full
- device_type
- dtb
- dtb_addr
- endian
- file_server_resource
- file_server_url
- initrd_addr
- job
- job_id
- kernel
- kernel_image
- kernel_image_size
- load_addr
- mach
- qemu
- qemu_command

gctucker: Oh, I thought this information was populated only once for each LAVA job in a "test suite" entry, so it shouldn't be duplicated.  Each test case in the suite then has a oid pointing to the suite.  Of course this is also in the boot entries, but that's normal as they're decoupled now.

The "test environment" is essentially a different combination of all these parameters for each LAVA job.  To save a bit of space, we could point to the build oid rather than copy fields from it (compiler, kernel_image, dtb, endian, defconfig).  Then the db queries would need to be adjusted but that seems pretty straightforward.

Might be worth some diagram.  Something like this essentially:

build
-----
* arch
* compiler
* dtb
* git_describe
* ...

test_suite
----------
* oid -> build
* name
* lab_name
* device_type
* boot_log
* ...

test_case
---------
* oid -> test_suite
* status
* time
...

Would this make any sense?

loysollivier: Yes makes sense. I didn't know what was your previous issue with boot/test duplicates. That's why I proposed to have a new document. But if it can just point to a build doc it's even better. 

I guess we could only keep lab_name and device_type in test suite. The dtb is tied to the build anyway ? 

Boot_log could be renamed test_suite_log (not necessarily now but in a future PR). And we could keep boot_log for the lava test suite ?

Essentially, one lava job reports one lava test_suite and potentially several other test_suites. So there is multiple test suites for each callback. Then it uses a structure to report the test elements which corresponds to the lava test structure: set and case.

gctucker: Ah OK, I think I didn't appreciate the part about multiple test suites in a single job as I've only been running simple jobs with only one test definition.  I'll add more test cases to my job template.  Still I should already have 2 suites: "lava" and a "simple" one with my current one - will take a closer look at this part.

Any further changes to the tables should probably be done in additional patches so we can see what differs from this version, which just mechanically adds fields from boot to test entries.

The dtb is actually part of the LAVA job data, not the build.  Yes the dtb files are coming from the build job, but it produces all the dtbs it can.  Then when submitting a job, only one dtb is picked (if applicable) for the target platform.  This is the piece of information we get from the callback.

About the log, well I just kept the same name as in boot entries i.e. `boot_log`.  Maybe we should just call it `log`?  The table name should make it obvious what kind of log it is, just like the `name` fields.

So I see what you mean about not duplicating data in several test suite entries.  Now, I think the common data should be in a different document.  We can't call it "job" as that's already used to call a category of tests (as shown on the front-end).  So here's what I mean, with a new "test" table which should have one entry per LAVA job (not strictly tied to LAVA though):

build
---
* arch
* compiler
* git_describe
* defconfig
...

test
---
* oid -> build
* lab_name
* device_type
* log
...

test_suite
---
* oid -> test
* name
...

test_case
---
* oid -> test_suite
* name
* status
...

Sorry this is getting a bit long...  I believe this should completely avoid duplication of data, i.e. things parsed from the callback should only be stored in one field in one table in the database.  It will probably make queries on the API front a bit indirect but I don't think this should be an issue; would be good to measure the performance impact though.

@MiloCasagrande Do you think it's a good idea to segment the data and avoid duplicates as in the scheme above?

loysollivier: Full ack on your latest comment @gctucker if we could start like that and send PRs to further changes that is good for me :)

MiloCasagrande: Usually I would advise against modeling the data in that way, but it depends.

The main user of the API and the data is going to be the frontend (but not only), so if it takes a lot of queries to retrieve the minimum data necessary for the frontend it's a no go. With data modeled in that way it becomes hard to answers questions like:
- What are the test cases associated with this build?
- What are the test cases associated with this tree?
- What are all the tests run by a single lab?

These are things that could be added to the frontend to extend it, or queries that we could run to do data analysis. With that data model, the backend would have to traverse at least 3 collections to retrieve the test cases ran (with a top-down approach even more with a bottom-up). If you embed or duplicate the data you have to traverse 1 collection. 

Although it might seem an anti-pattern, duplication or denormalization of data is actually welcomed in mongodb, and thinking about normalizing the data is an actual anti-pattern. The best approach when modeling the data with mongodb is: there shouldn't be any relations between collections, if there are, the data must be duplicated.

It's almost impossibile to not see relations between the various collections, but they should be kept at a minimum. It's possible to support queries like the questions above, but you might end up having to write really custom ones for each of the collections.

I wouldn't be concerned with space efficiency under this circumstances: it's not that we are duplicating massive amounts of data like parsed logs that can be megabytes in size . Documents are capped at 16MB, and our average document size is 1.9Kb (even the parsed logs summary average document size is way below the maximum level, and we are at 45Kb).

To support those kind of data layout, with documents linked through their `_id`, mongodb has a JOIN-like operator that performs a left-outer-join, but using it has its own costs: it's reliable and it's fast if you have massive indexes or if your data set is trivial, but the aggregation operations are not always that straightforward to write and the more collections you need to cross the harder it gets.

To be honest, I'm not so sure trying to normalize the data here is a good solution. Maybe some of the fields are not all that necessary in all the collections, but one of the thing discussed in the past was that we should have added more fields into the `test_set` and `test_case` collections to support more detailed queries (like on `job`, `kernel`, `defconfig`... values).

gctucker: @MiloCasagrande OK this is what I was wondering about, as mongodb is not a relational database with things linked to each other everywhere.  I was not suggesting that this should be exposed on the API though, but that the backend would put together the data with JOIN-like queries to return a flat structure to the client in one API call.

Found this part of the "little mongodb book" particularly relevant here:
https://github.com/karlseguin/the-little-mongodb-book/blob/master/en/mongodb.markdown#denormalization

Now it feels like we need a list of pros/cons reasons to make a good decision about splitting the tables or copying the data...  And in any case, we should review which fields we actually care about.  I already did a bit of this with the BootDocument, I only copied the ones that appeared to be actually used anywhere in the backend or frontend into the TestSuiteDocument with this PR.

loysollivier: @MiloCasagrande thanks for your input. I must admit, your comment leaves me a bit puzzled.
I am not a front end, database developer so I might miss some basic concepts for you guys.

But when I started working on the test documents structure in the database; I noticed that they are organized in a structured way: case belongs to sets that belongs to suites. 
If one would want to see in the frontend:
- all the test cases belonging to a test suite and test set.
He would need to find all the case and sets IDs in the suite document, and then search for this IDs in the DB. So it's already a three query approach where you could only do one if all the elements where in one single document.

Starting from this structure I extended it with more data. Now that we want to decouple boots from tests I thought it makes sense to create a "test" document that would have references to the test suites. This would be the exact same architecture than suites, sets and cases... 
Hence why I am a bit lost. What differs now ? You are suggesting that this isn't the correct approach but this approach was used to model the test structure at first. Why change now ?

Also, I am not against your approach, if duplicating data doesn't have a drawback on performance or space, why not just creating one big test document that contains everything ? This document would have all the test information from the test platform, build, names to all it's test suites, test sets and test cases results embedded. It would also be a lot easier on the frontend as there would be just one query on one big test document....

Also, using todays DB and proposed architecture, to answer your questions:
- What are the test cases associated with this build?
  - Now: From the build find all boots; from boots find all test suites; from test suites find all test cases
    - 3 queries
  - Proposition: From build find test; from test find suites; from suites find cases
    - 3 queries
- What are the test cases associated with this tree?
  - Now: search in build; from build find boots; from boot find suites; from suite find cases
    - 4 queries
  - Proposition: search in build; from build find test; from test find suites; find cases
    - 4 queries
Or maybe I am missing something and it is much easier today ? 

My point being:
- if it is denormalized : why not just create one big test document ?
- if there is some kind of normalization: why not replacing the boot document by a "test" document containing the same kind of information ?


Also an extract from @gctucker link on the "little monodb book":
_This doesn't mean you should duplicate every piece of information in every document. However, rather than letting fear of duplicate data drive your design decisions, consider modeling your data based on **what information belongs to what document.**_

The question is: what information belongs to what document ? :) 
IMHO I think that the test environment doesn't belong to a test suite, set or case that will be run on multiple platforms (and hence different test env). But I might be wrong, if you could enlighten me on that I'd be thankful !

MiloCasagrande:
On Wed, Dec 13, 2017 at 4:15 PM, Guillaume Tucker
<notifications@github.com> wrote:
>
> In app/models/test_suite.py:
>
> >              models.GIT_BRANCH_KEY: self.git_branch,
> +            models.GIT_COMMIT_KEY: self.git_commit,
>
> OK then I think we can just drop VCS_COMMIT_KEY from the test entries. I'll add another patch for that.

The vcs_commit was meant to point to the test definition commit, not
the git commit that is used on the boot report (the git kernel
commit).

If the test suite I'm running is hostend on a generic VCS, that should
point to the commit on that system. That was the idea for that field.

-- 
Milo Casagrande <milo@opensourcefoundries.com>

MiloCasagrande:
On Mon, Dec 18, 2017 at 10:50 AM, loys ollivier
<notifications@github.com> wrote:
>
> But when I started working on the test documents structure in the database;
> I noticed that they are organized in a structured way: case belongs to sets
> that belongs to suites.
> If one would want to see in the frontend:
>
> all the test cases belonging to a test suite and test set.
> He would need to find all the case and sets IDs in the suite document, and
> then search for this IDs in the DB. So it's already a three query approach
> where you could only do one if all the elements where in one single
> document.

It depends on how you want to visualize and explore the data. If you
want to visualize everything in one batch, yes, it's 3 database
queries and you are starting with zero prior knowledge of what you
want to look for.

It also depends a lot on how you want to define the user experience:
you can still show them everything in one single page, and although 3
queries might seem a lot, probably is not if you present the user with
initial information and then load the rest in the background.

The idea of separating them out was also made with a visualization in
mind: start from the test suite and then drill down to the single test
cases (and piggy-back the necessary information).

> Starting from this structure I extended it with more data. Now that we want
> to decouple boots from tests I thought it makes sense to create a "test"
> document that would have references to the test suites. This would be the
> exact same architecture than suites, sets and cases...
> Hence why I am a bit lost. What differs now ? You are suggesting that this
> isn't the correct approach but this approach was used to model the test
> structure at first. Why change now ?

Because we already have that in a boot document.

>
> Also, I am not against your approach, if duplicating data doesn't have a
> drawback on performance or space, why not just creating one big test
> document that contains everything ? This document would have all the test
> information from the test platform, build, names to all it's test suites,
> test sets and test cases results embedded. It would also be a lot easier on
> the frontend as there would be just one query on one big test document....

It's a compromise: you can do that, but then you'll always be
extracting the same big document even if all you need is a single test
case. You can "clean up" the final document in the response,
eliminating on the fly what is not necessary, but from a database
standpoint you are still extracting a huge document and then acting on
it. How big that document could be I have no idea, but with the
possibilities of running hundreds of tests and collecting data points,
that can grow easily big.

If you start having hundreds of embedded documents all into an array
contained in a single document, operations might also become more
problematic. Although write operations are atomic at the document
level, if you want to update a test case you'll need to traverse an
array (and the worst case scenario is that you don't have an index and
you have to scan it all).

Consider also that with the current way of thinking, you would have an
embedded document that contains other embedded documents (test suite
contains a list of test sets that each contain a list of test cases).

> Also, using todays DB and proposed architecture, to answer your questions:

Yes, with the actual DB data that is true, and if you want to see it
in SQL it is more or less already normalized, that is why I think we
would need to add even more data into each of the other documents
instead of extracting and normalizing it even further.

> What are the test cases associated with this build?
>
> Now: From the build find all boots; from boots find all test suites; from
> test suites find all test cases
>
> 3 queries
>
> Proposition: From build find test; from test find suites; from suites find
> cases
>
> 3 queries

Yep, you need to traverse 3 collections as I said, but if you add a
reference to the build to the test cases, it's 1 query and all you
have to access is 1 collection.

> What are the test cases associated with this tree?
>
> Now: search in build; from build find boots; from boot find suites; from
> suite find cases
>
> 4 queries
>
> Proposition: search in build; from build find test; from test find suites;
> find cases
>
> 4 queries
> Or maybe I am missing something and it is much easier today ?

Same as before: if you add a "job" field to the test suite and test
case, it's again 1 query.

>
> My point being:
>
> if it is denormalized : why not just create one big test document ?

See above.

> if there is some kind of normalization: why not replacing the boot document
> by a "test" document containing the same kind of information ?

And this is the part that for me has become again super confusing, and
I thought I understood the motives for this change: you are going to
create a "test environment" document that is 99.9% identical to a boot
document (linked to a build as well).

If it is a boot, why add another boot document (and collection) and
call it a test environment?

At that point the idea to just add a "filtering" field to identify a
boot used for testing is probably easier to implement (though it still
need the logic to understand when is one or the other at import time).

> Also an extract from @gctucker link on the "little monodb book":
> This doesn't mean you should duplicate every piece of information in every
> document. However, rather than letting fear of duplicate data drive your
> design decisions, consider modeling your data based on what information
> belongs to what document.

>
> The question is: what information belongs to what document ? :)

It could be all of it. Again, is a compromise.

> IMHO I think that the test environment doesn't belong to a test suite, set
> or case that will be run on multiple platforms (and hence different test
> env). But I might be wrong, if you could enlighten me on that I'd be
> thankful !

You are not wrong, but what you are talking about is a boot document,
and we already have that.

-- 
Milo Casagrande <milo@opensourcefoundries.com>

gctucker:
@MiloCasagrande Thanks for clarifying the `vcs_commit` field, not sure this is being used and documented correctly at the moment.  Will send an update to fix that.

@MiloCasagrande About using boot documents with tests, we could but it would mean changing a lot of code in the API implementation and possibly in the front-end as well to keep reporting boot test results in the same way.  This was discussed before, we ended up deciding to have a separate implementation for tests and eventually drop boot documents when they become superseded by tests.

>  IMHO I think that the test environment doesn't belong to a test suite, set
>  or case that will be run on multiple platforms (and hence different test
>  env). But I might be wrong, if you could enlighten me on that I'd be
>  thankful !

@loysollivier Each LAVA job will provide a "test environment" and some "test results" which will be linked together.  The environment will be the same for all the tests, so it's again a question of using a reference or a copy of this information.

> You are not wrong, but what you are talking about is a boot document,
> and we already have that.

@MiloCasagrande One issue with the current boot document and the related API and front-end usage is that it was all tailored around the concept of not doing anything else than a plain boot.  I hit this issue when trying to add the `test_plan` field in boot documents, I found quickly snowballed into a lot of changes at every level of the stack.


It seems like we should have a written description of what we're trying to achieve so we have a common document to refer to.  Maybe something to put in a Github issue?

MiloCasagrande:
On Mon, Dec 18, 2017 at 6:57 PM, Guillaume Tucker
<notifications@github.com> wrote:
>
> @MiloCasagrande Thanks for clarifying the vcs_commit field, not sure this is being used and documented correctly at the moment. Will send an update to fix that.

The API docs report a definition_uri and a vcs_commit fields with the
description of how they should be used. Descriptions could definitely
be improved though.
For the fact that is not used, I guessed nobody ever really used the
API to report the all the data it can include.

> @MiloCasagrande About using boot documents with tests, we could but it would mean changing a lot of code in the API implementation and possibly in the front-end as well to keep reporting boot test results in the same way. This was discussed before, we ended up deciding to have a separate implementation for tests and eventually drop boot documents when they become superseded by tests.

Code would need to be changed anyway, I'm not so sure it's really that
big of a change.

The boot report visualization is already in place, and you get to use
that as well: if we add more field to the boot data set, we just need
to show them.
You might want to have a few different things shown or not based on
the fact that a boot is of a different "type", but I don't think it's
that massive of a work (unless we talk about a complete overhaul of
the UI).

> IMHO I think that the test environment doesn't belong to a test suite, set
> or case that will be run on multiple platforms (and hence different test
> env). But I might be wrong, if you could enlighten me on that I'd be
> thankful !
>
> @loysollivier Each LAVA job will provide a "test environment" and some "test results" which will be linked together. The environment will be the same for all the tests, so it's again a question of using a reference or a copy of this information.

> @MiloCasagrande One issue with the current boot document and the related API and front-end usage is that it was all tailored around the concept of not doing anything else than a plain boot.

All the rest apart from a boot, should go into a test.
Corner case might be NFS boots or other different boot methods, but I
think it's easier to adapt the boot document we have now to store that
information rather than changing the entire data model.

> It seems like we should have a written description of what we're trying to achieve so we have a common document to refer to. Maybe something to put in a Github issue?

Let's use a shared Google docs or something we can all edit at once,
and let's also start from the beginning of the problem:

https://docs.google.com/document/d/1crbDDlYL8HJZXlbLQtfH6uUpw0zZCZq-5zm2ol7S61U/edit?usp=sharing

It's in view mode only, please request write access.

-- 
Milo Casagrande <milo@opensourcefoundries.com>

gctucker:
> let's also start from the beginning of the problem
I thought we had already spent the time needed to reach an agreement on the way to proceed with this rework, i.e. don't touch the boot results, extend the test results independently and when they're mature enough switch to use only test results.  I guess we need to restart this train of thought.

MiloCasagrande:
On Tue, Dec 19, 2017 at 9:33 AM, Guillaume Tucker
<notifications@github.com> wrote:
>
> I thought we had already spent the time needed to reach an agreement on the way to proceed with this rework, i.e. don't touch the boot results, extend the test results independently and when they're mature enough switch to use only test results. I guess we need to restart this train of thought.

Well, if we start talking about extracting the "test environment" data
(that is the boot data) into a separate collection, then I don't
understand anymore why we just cannot use the already available boot
report, especially when saying that one will just disappear in the
future.

-- 
Milo Casagrande <milo@opensourcefoundries.com>

gctucker:
Thanks @MiloCasagrande for the discussion.  So this pull request is only to separate the test results from boots, to solve the current known problems without impacting the current functionality around boot testing.  We can decide to reshape the database and the backend code later on if need be.

MiloCasagrande:
@gctucker I'm going to merge this into the staging branch first and deploy it on staging. I don't plan to update production until the end of the week so it gives us a few days to test it out there.

MiloCasagrande:
This has been deployed on staging.

PR #47, opened 2017-12-19
utils/callback: lava.py store raw json data from LAVA callbacks

loysollivier:
Add a utility that stores the raw test data from a LAVA v2 callback.

It would be very usefull to have this kind of data stored along with the
logs. This can help debug issues happening on the frontend / backend by
replaying them. It can also help in development to fake callback events
and make sure there was no regression.

The raw json data from LAVA callback is stored in a folder called
"json-raw", with the default PATH being the utils.BASE_PATH.

Beforing storing the data, sensitive information such as the token used
to send the callback is removed. A link to the boot_log, lab_name and
version is also added for further use.

### DIFF
--- a/app/utils/callback/lava.py
+++ b/app/utils/callback/lava.py
@@ -15,6 +15,7 @@ import errno
 import models
 import os
 import yaml
+import json
 
 import utils
 import utils.boot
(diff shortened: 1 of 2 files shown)


loysollivier:
@MiloCasagrande would it be possible to integrate this change in kernelci ? It would be very useful for me to have on storage.kernelci... these json files.

I created them in a way such as they do not contain any sensitive information. And using this I could replicate the database of kernelci locally for my dev and tests.

app/utils/callback/lava.py
```diff
@@ -235,6 +236,36 @@ def _add_boot_log(meta, log, base_path):
...
+    :param job_data: The JSON data from the callback.
+    :type job_data: dict
+    :param base_path: Path to the top-level directory where to store the files.
+    :type base_path: string
+    """
+
+    utils.LOG.info("Saving raw LAVA v2 callback data in {}".format(base_path))
+    job_definition = yaml.load(job_data["definition"], Loader=yaml.CLoader)
+    file_name = job_definition["job_name"] + ".json"
+    file_path = os.path.join(base_path, "json-raw", file_name)
```
MiloCasagrande: My only concern here is with the filesystem path. Why not storing them at the same path where the boot log (HTML and TXT) files are stored?

I understand is more practical to get all the JSON files from one directory, but we also have to deal with that directory on server side and clean it up from time to time (if the file is stored along with the boot logs, it gets clean automatically after 40 days).

MiloCasagrande: BTW, a long term plan (briefly discussed) is to open the data base data and release it publicly (like every 6 months). It's still up for discussion, we don't know the license etc... but it could also solve part of the problem of grabbing some data to work with.

loysollivier: OK I will move it to be stored along the boot logs. 

Getting the raw JSON would already be a nice first step in sharing the data :)
Restoring the database would also be useful, we could use it with docker then.

loysollivier:
@MiloCasagrande edited and rebased with your changes request

loysollivier:
Cool ! Can you already tell when this changes will reach staging or kernelci.org ? So I can look out for the files :)

Thanks

MiloCasagrande:
On Fri, Dec 22, 2017 at 9:06 AM, loys ollivier <notifications@github.com> wrote:
>
> Cool ! Can you already tell when this changes will reach staging or kernelci.org ? So I can look out for the files :)

I'll try and deploy it next week at least on staging (if somebody else
doesn't beat me to it).
Production will have to wait until January I guess: I'm traveling for
the next couple of weeks.

Ciao.

-- 
Milo Casagrande <milo@opensourcefoundries.com>

MiloCasagrande:
I deployed the changes on staging just now.
Ciao.

loysollivier:
Thanks a lot !
Safe travel & happy holidays :)

PR #50, opened 2018-01-30
TestSuite: fix new document ids

gctucker:
New document IDs are only generated when there is no ID key at all in
the new object's dictionary when passed to utils.db.save() - having a
key set to None results in a null key in the database.  Fix this by
reverting to only adding an ID key when the object already has a valid
ID.

Fixes: 21f1eb56e714 ("Add boot meta-data to TestSuiteDocument")

### DIFF
--- a/app/models/test_suite.py
+++ b/app/models/test_suite.py
@@ -147,7 +147,7 @@ class TestSuiteDocument(modb.BaseDocument):
         self._created_on = value
 
     def to_dict(self):
-        return {
+        test_suite = {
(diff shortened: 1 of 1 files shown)


gctucker:
Job showing the issue:
https://staging.kernelci.org/test/board/qemu/job/gtucker/kernel/kernelci-test-053-4-g66ec089-lava-bisect-v2-b-staging-5091-3/

Job run with this branch showing the issue is fixed:
https://staging.kernelci.org/test/board/qemu/job/gtucker/kernel/kernelci-test-054/

PR #55, opened 2018-02-28
Test documents: store raw json logs as well

loysollivier:
Since the boot and test documents have been splitted, the raw json would
only be recorded if calling on the /boot/ api.

Also record the json logs if it is a test callback.

### DIFF
--- a/app/utils/callback/lava.py
+++ b/app/utils/callback/lava.py
@@ -486,4 +486,6 @@ def add_tests(job_data, lab_name, db_options, base_path=utils.BASE_PATH):
             if errors:
                 raise utils.errors.BackendError(errors)
 
+    store_lava_json(job_data, suite_data)
+
     return suite_doc_id


khilman:
Validated on the kernelci-docker install at kernelci.dev.baylibre.com:8081

app/utils/callback/lava.py
```diff
@@ -296,6 +296,7 @@ def store_lava_json(job_data, meta, base_path=utils.BASE_PATH):
     job_data["lab_name"] = meta.get("lab_name")
     job_data["version"] = meta.get("version")
     job_data["boot_log_html"] = meta.get("boot_log_html")
+    job_data["test_log_html"] = meta.get("log")
```
gctucker: `meta["log"]` is not HTML, it's a YAML log coming directly from LAVA.  So it should probably be called something else in `job_data`, such as `job_data["log"]` or `job_data["lava_log"]`.

app/utils/callback/lava.py
```diff
@@ -486,4 +487,6 @@ def add_tests(job_data, lab_name, db_options, base_path=utils.BASE_PATH):
             if errors:
                 raise utils.errors.BackendError(errors)
 
+        store_lava_json(job_data, suite_data)
```
gctucker: It seems to me that if you put this in the loop that iterates over all the test suites, the same JSON file will be created each time but in a different directory with the suite name.  Should this not be called only once outside the loop, like for boots?

Also, keep in mind that `add_test` is called for every callback received, so boots will have their JSON data stored twice with this approach.  As I don't think this was ever actually used for boot data in the past, being a new thing with tests it could be only called in `add_tests` and not in `add_boot`.

loysollivier:
Even simpler, this becomes a one liner. The log is already in the test document with the correct name. And indeed one call at the end of the import is enough.

loysollivier:
Also, we would like to get both the boot and test data for different test purposes. I would propose that we keep it in both callbacks. IIRC this should not be duplicated, the callback happens either on boot or test but not on both.

loysollivier:
Hello guys any comments on this ?

MiloCasagrande:
Apologies this took so long, it sort of slipped out of my view.
Merging.

PR #59, opened 2018-06-05
T7884 Add handler for /bisect POST to save bisection results

gctucker:
Add a BisectHandler._post() method to handle POST requests and run a
task to update a BootBisectionDocument with the results.  This relies
on having a document previously created when triggering the bisection.

### DIFF
--- a/app/handlers/bisect.py
+++ b/app/handlers/bisect.py
@@ -20,6 +20,7 @@ import handlers.base as hbase
 import handlers.common.query
 import handlers.response as hresponse
 import models
+import taskqueue.tasks
 import taskqueue.tasks.bisect as taskt
 import utils.db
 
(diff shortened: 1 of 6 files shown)


MiloCasagrande:
Seems ok to me, do you want to test it on staging before?
We can merge and test the master branch directly before rolling out to kci.org.

gctucker:
Thanks @MiloCasagrande, yes I want to test it first together with some changes in `lava-ci`.  Then there's another change to send bisection email reports, and once we've got all that I think it will be time to roll it out on production.

gctucker:
I've tested it OK, just had to change `kernelci-celery.cfg` to allow POST requests on `/bisect`: https://github.com/kernelci/kernelci-backend-config/pull/20

MiloCasagrande:
Sounds good. I'll merge it and then we can decide when to roll it out.

gctucker:
Great, thanks.  I'm triggering another round of build/boot/bisect on staging with emails now, to verify it's all working as expected.  We'll probably need to do a final run again with the master branches on staging when it's all merged.

PR #62, opened 2018-06-18
T9475 Add BisectDocument.kernel with kernel identifier

gctucker:
Rather than relying on the output of "git describe" to consistently
track the kernel name, keep the string originally used.  This solves
issues with various versions of Git showing different default output
format for "git describe", and makes the logic more future-proof in
case kernel identifiers are not plain "git describe" outputs any more.


MiloCasagrande:
Looks good to me.

gctucker:
Tested OK on staging, merging now.

PR #57, opened 2018-05-24
Decode log lines in the LAVA callback to utf8

mattface:
Recent python3 LAVA versions send the log files encoded with utf8. The API needs to decode these in order to use them correctly.

This also needs testing with an older LAVA lab as well

PR #63, opened 2018-06-25
T9356 lava_log_parser: write logs as Unicode

gctucker:
Handle incoming Unicode strings by opening the log files in Unicode
mode and not converting log lines to ASCII.

With LAVA moving to Python 3 which uses Unicode for all the strings,
some characters were not compatible with ASCII any more, causing some
logs to fail to be parsed.  This fixes the issue and remains
compatible with older versions of LAVA running Python 2.7.

### DIFF
--- a/app/utils/callback/lava.py
+++ b/app/utils/callback/lava.py
@@ -11,6 +11,7 @@
 # You should have received a copy of the GNU Affero General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+import codecs
 import errno
 import models
 import os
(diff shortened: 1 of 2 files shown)


gctucker:
This replaces PR #57 which didn't quite solve the problem.

mattface:
Has this been tested on staging?

gctucker:
Tested on staging, there are no errors in the backend log.  Here's an example of a boot with non-ASCII characters that got parsed and converted to HTML successfully:
http://staging-storage.kernelci.org/gtucker/kernelci-test/kernelci-test-066-8-g1fe9fc17a9c7/x86/x86_64_defconfig/lab-mhart-dev/boot-x86-atom330.html

PR #66, opened 2018-07-04
T9718 build.txt: fix Unicode handling errors

gctucker:
The Jinja2 .format() filter automatically treats the template string
as ASCII unless it contained Unicode characters, on Python 2.7.  If
the values passed to this template are Unicode, then it raises a
UnicodeEncodeError as it can't convert Unicode to ASCII.

Use the % format in build.txt Jinja2 template to avoid this issue.
This would probably not be an issue on Python 3 which treats all
strings as Unicode.  The build.txt template was the only one to show
this problem.

### DIFF
--- a/app/utils/report/templates/build.txt
+++ b/app/utils/report/templates/build.txt
@@ -46,19 +46,19 @@
 {%- if errors_summary.errors %}
 Errors summary:
 {% for err in errors_summary.errors %}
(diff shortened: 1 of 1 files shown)


gctucker:
Some background about this issue:
https://github.com/CERNDocumentServer/cds-videos/commit/739845cae3ac12be4b70cccb8fc71f552a87ece5

PR #70, opened 2018-08-08
Send report emails about test suites

ana:
This is a first approach to send emails with the result of the test suites. These reports will be improved after some testing, specially now that we should be having more results in the database. For the moment, they only allow to trigger emails for reporting all the tests ran for the same tree+branch+kernel, like this:

```
curl -X POST -H "Authorization: $EMAIL_AUTH_TOKEN" -H "Content-Type: application/json" -d '{"job": "'$TREE_NAME'", "kernel": "'$GIT_DESCRIBE'", "git_branch": "'$BRANCH'", "report_type": "test", "format": ["txt"], "send_to": ["ana@collabora.com"], "delay": 300}' ${API}/send
```

The emails generated have the following format:

```
Subject: Test suite result for local/master -  v4.17-6625-g1c8c5a9d38f6

Test suite result for local/master  - v4.17-6625-g1c8c5a9d38f6

Testsuite: lava

  *  Tree:        local
  *  Branch:      master
  *  URL:         /home/ana/KernelCI/src/linux
  *  Kernel:      v4.17-6625-g1c8c5a9d38f6
  *  Config:      defconfig
  *  Config Full: defconfig
  *  Lab Name:    lab-agl
  *  Board:       qemu
  *  Git Commit:  1c8c5a9d38f607c0b6fd12c91cbe1a4418762a21
  *  Date:        2018-06-11 00:07:32.196000

  * Test cases:

    * job : PASS
    * auto-login-action : PASS
    * http-download-ramdisk : PASS
    * http-download-kernel : PASS
    * validate : PASS

Testsuite: lava

  *  Tree:        local
  *  Branch:      master
  *  URL:         /home/ana/KernelCI/src/linux
  *  Kernel:      v4.17-6625-g1c8c5a9d38f6
  *  Config:      defconfig
  *  Config Full: defconfig
  *  Lab Name:    lab-agl
  *  Board:       qemu
  *  Git Commit:  1c8c5a9d38f607c0b6fd12c91cbe1a4418762a21
  *  Date:        2018-06-11 13:17:00.531000

  * Test cases:

    * validate : PASS
    * http-download-ramdisk : PASS
    * http-download-kernel : PASS
    * test-overlay-v4l2 : PASS
    * test-install-overlay-v4l2 : PASS
    * test-runscript-overlay-v4l2 : PASS
    * apply-overlay-guest : PASS
    * auto-login-action : PASS
    * 0_v4l2 : PASS
    * job : PASS

Testsuite: v4l2

  *  Tree:        local
  *  Branch:      master
  *  URL:         /home/ana/KernelCI/src/linux
  *  Kernel:      v4.17-6625-g1c8c5a9d38f6
  *  Config:      defconfig
  *  Config Full: defconfig
  *  Lab Name:    lab-agl
  *  Board:       qemu
  *  Git Commit:  1c8c5a9d38f607c0b6fd12c91cbe1a4418762a21
  *  Date:        2018-06-11 13:17:00.402000

  * Test cases:

    * v4l2-compliance : FAIL

```

### DIFF
--- a/app/handlers/send.py
+++ b/app/handlers/send.py
@@ -11,7 +11,7 @@
 # You should have received a copy of the GNU Affero General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-"""This module is used to send boot and build report email."""
+"""This module is used to send boot, build and test report emails."""
 
 import bson
 import copy
@@ -57,6 +57,11 @@ REPORT_TYPE_KEYS = {
         models.DEVICE_TYPE_KEY,
         models.BISECT_GOOD_COMMIT_KEY,
         models.BISECT_BAD_COMMIT_KEY,
+    ],
+    'test': [
+        models.JOB_KEY,
+        models.GIT_BRANCH_KEY,
+        models.KERNEL_KEY,
     ],
 }
 
@@ -338,6 +343,48 @@ class SendHandler(hbase.BaseHandler):
             error_string = (
                 "No email addresses provided to send bisection report to")
             self.log.error(error_string)
+
+        return has_errors, error_string
+
+    def _schedule_test_report(self, report_data, email_opts, countdown):
+        """Schedule the test report performing some checks on the emails.
+
+        :param report_data: Contents to use in the report.
+        :type report_data: dict
+        :param email_opts: The data necessary for scheduling a report.
+        :type email_opts: dictionary
+        :param countdown: Delay time before sending the email.
+        :type countdown: int
+        :return A tuple with as first parameter a bool indicating if the
+        scheduling had success, as second argument the error string in case
+        of error or None.
+        """
+        has_errors = False
+        error_string = None
+
+        job, git_branch, kernel = (report_data[k] for k in [
+            models.JOB_KEY,
+            models.GIT_BRANCH_KEY,
+            models.KERNEL_KEY,
+        ])
+
+        if email_opts.get("to"):
+            taskq.send_test_report.apply_async(
+                [
+                    job,
+                    git_branch,
+                    kernel,
+                    report_data,
+                    email_opts,
+                ],
+                countdown=countdown
+            )
+        else:
+            has_errors = True
+            error_string = "No email addresses provided to send test report to"
+            self.log.error(
+                "No email addresses to send test report to for '%s-%s-%s'",
+                job, git_branch, kernel)
 
         return has_errors, error_string
 
--- a/app/taskqueue/tasks/report.py
+++ b/app/taskqueue/tasks/report.py
@@ -22,6 +22,7 @@ import utils.report.boot
 import utils.report.build
 import utils.report.common
 import utils.report.error
+import utils.report.test
 
 
 # pylint: disable=too-many-arguments
@@ -172,6 +173,42 @@ def send_bisect_report(report_data, email_opts, base_path=utils.BASE_PATH):
 
     utils.report.common.save_report(
         job, git_branch, kernel, models.BISECT_REPORT,
+        status, errors, db_options)
+
+    return status
+
+
+@taskc.app.task(name="send-test-report")
+def send_test_report(job, git_branch, kernel, report_data, email_opts):
+    """Send the test suite report email.
+
+    :param job: The job name.
+    :type job: string
+    :param git_branch: The git branch name.
+    :type git_branch: string
+    :param kernel: The kernel name.
+    :type kernel: string
+    :param report_data: The data necessary for generating a report.
+    :type report_data: dict
+    :param email_opts: Email options.
+    :type email_opts: dict
+    """
+    report_id = "-".join([job, git_branch, kernel])
+    utils.LOG.info(
+        "Sending test suite report email for '{}'".format(report_id))
+
+    db_options = taskc.app.conf.get("db_options", {})
+
+    txt_body, html_body, subject, headers = \
+        utils.report.test.create_test_report(
+            report_data, email_opts["format"], db_options)
+
+    status, errors = utils.emails.send_email(
+        subject, txt_body, html_body, email_opts,
+        taskc.app.conf.mail_options, headers)
+
+    utils.report.common.save_report(
+        job, git_branch, kernel, models.TEST_REPORT,
         status, errors, db_options)
 
     return status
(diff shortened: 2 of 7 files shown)


khilman:
This looks like a good start.
Curious why your example has to sections for "Testsuite: lava"

Also, it would be nice to have a high-level summary at the Testsuite level.  e.g. something like
Testsuite: lava   N tests, Pass rate M% (X PASS, Y FAIL, Z SKIP)

ana:
@khilman 
> This looks like a good start.
> Curious why your example has to sections for "Testsuite: lava"
That was a mistake

ana:
I have updated the PR with the following changes:
- Remove the common data from every testsuite (tree, branch, kernel, git_url, git_commit), they're only shown once on the top of the mail now
- Do not show all the "lava" test suites, they're the build test and the kselftest tests. The kselftests should be updated to start appearing as "kselftest" test suites in the database.
- Add the stats lines with the number of total tests, passed, failed and skip tests

Example mail for  khilman/to-build - v4.18-rc8-53-gb4a2f821c4a1
https://gist.github.com/ana/af7f153384e95588132c588dac983215

And for stable/linux-4.14.y - v4.14.62
https://gist.github.com/ana/f5ba07193329234fdfd0f40a4f0f019c

ana:
hi! any comments or suggestion on this PR?
The reports are a bit long, but this is the first version. I'm planning to add some improvements to also allow sending the reports filtered by lab/board/test. Right now, I'd like to be able to start getting reports for the current test suites.

khilman:
IMO, Let's merge this and then iterate/improve.

Some thoughts:

It's missing a high-level summary at the top.  e.g. how many test suites ran?  how many passed/failed?  on what boards?

For example, something like for each test suite ran, how many sets/cases pass/fail/skip for each, and how many boards were each run on.  The idea is that from the first page of the email report, you have a pretty good idea whether most things are passing/failing without having to scroll into the details.


Suites    Sets   Cases   Boards
------------------------------------------------------
sleep:     n/a     X/Y/Z      3
usb
simple
cros-ec
igt
v4l2

app/handlers/send.py
```diff
@@ -11,7 +11,7 @@
 # You should have received a copy of the GNU Affero General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-"""This module is used to send boot and build report email."""
+"""This module is used to send boot, build and test suite report emails."""
```
gctucker: and also bisections... Maybe this could just say "to send report emails".

app/taskqueue/tasks/report.py
```diff
@@ -177,6 +178,41 @@ def send_bisect_report(report_data, email_opts, base_path=utils.BASE_PATH):
...
+    :param kernel: The kernel name.
+    :type kernel: string
+    :param report_data: The data necessary for generating a report.
+    :type report_data: dict
+    :param email_opts: Email options.
+    :type email_opts: dict
+    """
+    report_id = "-".join([job, git_branch, kernel])
+    utils.LOG.info("Sending test suite report email for '{}'".format(report_id))
+    status = "ERROR"
```
gctucker: Not strictly required as this gets initialised when calling `utils.emails.send_email()`.

app/utils/report/templates/test.html
```diff
@@ -0,0 +1,63 @@
...
+          <tr><th>Lab Name</th><td>{{ t.lab_name }}</td></tr>
+          <tr><th>Board</th><td>{{ t.board }}</td></tr>
+          <tr><th>Date</th><td>{{ t.created_on }}</td></tr>
+          <tr><th>Test cases:</th><td></td></tr>
+          <tr><th></th><td>
+          <table>
+            {% for tc in t.test_case_list %}
+            <tr><th>{{ tc.name }}:</th><td>{{ tc.status }}</td></tr>
+            {%- endfor %}
+            </table>
```
gctucker: Small indentation glitch :)

app/utils/report/templates/test.html
```diff
@@ -0,0 +1,63 @@
+<!doctype html>
```
gctucker: Do you have an example of what the HTML version of the report looks like?

app/utils/report/test.py
```diff
@@ -0,0 +1,149 @@
+# This program is free software: you can redistribute it and/or modify
```
gctucker: We should also have a `utils.report.tests.test_test_report` unit test to cover this new module (and one for bisections as well I guess).

app/utils/report/test.py
```diff
@@ -0,0 +1,149 @@
...
+               {"_id": tc})
+            ts['test_case_list'].append(test_case_doc)
+
+        resultlist = [test['status'] for test in ts['test_case_list']]
+        ts['total_tests'] = len(resultlist)
+        ts['total_pass'] = len([e for e in resultlist if e=="PASS"])
+        ts['total_fail'] = len([e for e in resultlist if e=="FAIL"])
+        ts['total_skip'] = len([e for e in resultlist if e=="SKIP"])
+
+    # Remove all the "lava" test suites, they're the boot tests
```
gctucker: It's not quite accurate that the `lava` test suites are the boot tests.  Rather, boot tests only have a `lava` test suite.  Ultimately it would be good to deal with failing boot results for any test plan (say, report that a `usb` test plan failed to complete) as that would also include the plain `boot` test plan without artificially creating a test suite for it.

For the comment here, I think it's better to say that the `lava` test suite doesn't contain any actual results from the defined test plan.

app/utils/report/test.py
gctucker: Also, does this pass pep8 and existing unit tests?

ana:
I just pushed v3 addressing the remarks from @gctucker. For the unit test part, the coverage is the same than master and adding `utils.report.tests.test_test_report` is still pending.
I also made some improvements in the templates, currently the emails looks like:
 *  txt: https://gist.github.com/ana/a02b85909985c5c21fe75be83826a9c8
*  html: https://gist.github.com/ana/d17fd6d5609b6616a166e33ae16573aa (you'll need to safe the raw gist as mail.html and read it with your browser)

ana:
@khilman @gctucker  the HTML was mangled by thunderbird, I have updated the gist and if you save as a html now it should look good :)
About the colors, none of the HTML reports right now have any, if somebody wants to write a nice CSS we can colour of all of them at the same time :)

gctucker:
Thanks, looks good to me.

khilman:
Looks good now, thanks.

PR #71, opened 2018-08-15
[bugfix] templates/multiple_emails.txt: fix email_type

ana:
The template multiple_emails.txt was expecting for a list of strings
(report type) in the variable email_type, but this variable is
a single string with the type (build, boot, test).


MiloCasagrande:
Looks good, thanks!

PR #72, opened 2018-08-23
add image.type to test_suite document

khilman:
Parse image.type from the kernelCI metadata and save into the
test_suite document stored in the db.

All kernelCI jobs are setting this field to "kernel-ci" today, but we
could use this to distinguish different types of images (e.g. yocto,
etc.)


MiloCasagrande:
Looks good to me.

PR #73, opened 2018-08-31
Fixes for pep8  and unit test

ana:
Two commits:
- fixing a couple of long standing pep8 errors
- updating the test suites following https://github.com/kernelci/kernelci-backend/pull/72

### DIFF
--- a/app/utils/callback/lava.py
+++ b/app/utils/callback/lava.py
@@ -256,9 +256,9 @@ def _add_boot_log(meta, job_log, base_path, name):
             if e.errno != errno.EEXIST:
                 raise e
 
-    with codecs.open(txt_path, "w", "utf-8") as txt, \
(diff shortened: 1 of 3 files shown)


app/utils/callback/lava.py
```diff
@@ -256,9 +256,9 @@ def _add_boot_log(meta, job_log, base_path, name):
             if e.errno != errno.EEXIST:
                 raise e
 
-    with codecs.open(txt_path, "w", "utf-8") as txt, \
```
gctucker: Alternative using only one `with`:
```
    with (codecs.open(txt_path, "w", "utf-8"),
          codecs.open(html_path, "w", "utf-8")) as txt, html:
        utils.lava_log_parser.run(log, meta, txt, html)
```
but using 2 nested `with` statements is fine too.

PR #74, opened 2018-09-03
Remove test set

ana:
Remove fully the current test set implementation from the backend.
This shouldn't affect the test result data stored in the backend because all the information is stored in the test suite collection as well.
This is a first step in the new implementation on how test results are stored in the backend. 

Frontend PR with corresponding changes: https://github.com/kernelci/kernelci-frontend/pull/34


gctucker:
Tested OK on staging.

PR #75, opened 2018-09-06
WIP T10505 test groups

gctucker:
Implement test groups with test results, and nested sub-groups.  It depends on #74 to be merged first.

This is an early PR for review.  It can:
* parse LAVA test callback data into test groups with test cases, and test sets as sub-groups
* recursively import into the database a structure with test groups containing sub-groups and cases
* generate test email reports (text and html) with one level of sub-groups

What it still needs to do:
* accept test data from the HTTP POST API request (i.e. not the LAVA callback entry point)
* update documentation and database schema documents


gctucker:
Sample test report with sub-groups:
https://gist.github.com/gctucker/01b3584739606512b1a37eaeedf33410

gctucker:
I found this while working on the code, I guess it could be moved to a separate PR as it's not actually related to test groups.  I also don't think there should be any missing arch data from the callback data, it would like a missing lab name or device type or kernel version imo.

gctucker:
And to answer your question in particular (sorry), no it won't dump stack but return an HTTP error.  If there's no arch set, the `.get()` call will return `None` and the check with `VALID_ARCHITECTURES` will fail.

khilman:
OK, thanks for clarifying.  I mistakenly thought .get() would fail if you don't provide it with the "missing" value.

gctucker:
I've moved the default arch fix to #76.

ana:
so far, looks good to me!

gctucker:
Sample email test reports with modified "simple" test plan to include sets:
https://people.collabora.com/~gtucker/tmp/test-report-email-002.html
https://people.collabora.com/~gtucker/tmp/test-report-email-002.txt

PR #80, opened 2018-10-15
report/templates/test: add board and arch to each detailed section

gctucker:
Repeat the board name and architecture from the summary in each
detailed section to make it easier to follow.


gctucker:
@ana That's right, also the test results are already there in the template a bit further down.
test-results-email-html-thunderbird

PR #81, opened 2018-10-16
Allow to override DEFAULT_STORAGE_URL and DEFAULT_BASE_URL

ana:
When using test instances, it's useful to have the mail reports showing the url from the local instance. This change allows to override these values in the file `/etc/linaro/kernelci-backend.cfg` using the values `base_url `and `storage_url`

### DIFF
--- a/app/utils/report/common.py
+++ b/app/utils/report/common.py
@@ -44,9 +44,9 @@ JOB_SEARCH_FIELDS = [
 ]
 
 DEFAULT_BASE_URL = u"https://kernelci.org"
-DEFAULT_BOOT_URL = u"https://kernelci.org/boot/all/job"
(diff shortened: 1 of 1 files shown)


gctucker:
@ana Could you please give an example of which config values need to be used on staging?  The `base_url` isn't set there at the moment, only `storage_url` is set to `http://storage.kernelci.org/`.

ana:
> http://storage.kernelci.org/



> @ana Could you please give an example of which config values need to be used on staging? The `base_url` isn't set there at the moment, only `storage_url` is set to `http://storage.kernelci.org/`.

`storage_url` should be updated to `http://staging-storage.kernelci.org/` and `base_url` should be `https://staging.kernelci.org/`

gctucker:
Thanks, tested OK on staging.

PR #83, opened 2018-11-08
Update test reports with the build_data.json information from the rootfs

ana:
The rootfs images provide information about their content in a JSON file named build_info.json stored in the same web directory that the rootfs.
This PR downloads the build_info.json file, if it exists, and imports it in the backend/mongodb every time a new test_group document is added.
The test reports templates have been also updated to add this information.


gctucker:
@ana What if a test doesn't use any initrd but has its rootfs say, via NFS or on a local storage medium?

ana:
> @ana What if a test doesn't use any initrd but has its rootfs say, via NFS or on a local storage medium?

If `initrd_url` is not returned by the LAVA callback, it would have been showing an empty field in the report. I have just updated the PR to not show the rootfs line in this case. 
I'd need to investigate what the LAVA callback is returning if the test is using NFS / local medium for handling it properly.

ana:
I just updated the last commit to fix a typo in the HTML template.

app/utils/callback/lava.py
```diff
@@ -457,6 +458,33 @@ def _add_test_results(group, suite_results, suite_name):
...
+    """Add rootfs info
+
+    Download the the json file with the information of the rootfs if it's
+    available and add its information to the group data
+
+    :param group: Test group data.
+    :type group: dict
+    """
+
+    rootfs_url = group["initrd"]
```
gctucker: Note: This would fail in the case where a job ran without any `initrd` such as the current `boot-nfs` test plan.  However, we're lucky as the `boot-nfs` test plan doesn't actually have any test cases so this function isn't called.

For the time being, I think it would be safer to do this:
```
rootfs_url = group.get('initrd')
if not rootfs_url:
    return
```

Then later on, add support for other rootfs types i.e. nfs...

ana: Updated and tested locally. Thanks!

PR #86, opened 2018-11-12
Add a build environment required field to builds

mattface:
In order to differentiate between builds with different compilers, or "build environments" in docker, add a required field to the backend when uploading a build.

Also, add compiler version to the defconfig build result emails to show the differences.


mattface:
Relies on https://github.com/kernelci/kernelci-core-staging/pull/43

MiloCasagrande:
Changes look good.

I guess this will then need to be propagated to the frontened as well, since a few of those links will need to be updated.

mattface:
I'm just going to clarify what I'm trying to achieve with this on the mailing list, at request of Guillaume

app/models/build.py
```diff
@@ -38,6 +38,8 @@ def __init__(
         :param defconfig_full: The full value of the defconfig when it contains
         fragments. Default to the same 'defconfig' value.
         :type defconfig_full: string
+        :param build_environment: The docker image used to build the kernel_image
```
gctucker: Rather than explicitly mentioning Docker, it may be better to say something more generic such as "description of the build environment used to build the kernel_image" or something like that.  We may not always use Docker, and it would be good if this field could still be used with other build environments.

mattface: Done

gctucker:
Also please rebase on master and fix / update the unit tests.

mattface:
rebased and unit tests fixed

mattface:
This should not be merged yet, as it requires 1 more patch to deal with boot artifacts overwriting each other. We will need to use the file_server_resource of the linked build to make sure the log files are written to the correct place.
I have a working patch for this and will be adding it to this PR.

gctucker:
So with that extra fix, will the backend changes need to be merged before or at the same time as the kernelci-core change?

mattface:
This has been rebased to include the recent changes to architecture keys and is ready to be merged, with https://github.com/kernelci/kernelci-core-staging/pull/66

mattface:
@gtucker it should be mandatory, and yes it's a bug in build.py to not ensure they are set.
It's the same problem with GIT_DESCRIBE for example, if you dont set that build.py just blows up with an error. I agree with the fix from https://github.com/kernelci/kernelci-core-staging/pull/66, I'll make the environment required and also overrideable.

app/utils/report/templates/build.html
```diff
@@ -163,7 +163,7 @@
         {%- set warns = P_("{:d} warning", "{:d} warnings", d.warnings_count).format(d.warnings_count) %}
         {%- set mism = P_("{:d} section mismatch", "{:d} section mismatches", d.mismatches_count).format(d.mismatches_count) %}
         <p>
-    {{ "<strong>{}</strong> ({}) &mdash; {}, {}, {}, {}".format(d.defconfig_full, d.arch, d.status, errs, warns, mism) }}
+    {{ "<strong>{}</strong> ({}, {}) &mdash; {}, {}, {}, {}".format(d.defconfig_full, d.arch, d.compiler_version_ext, d.status, errs, warns, mism) }}
```
gctucker: I keep getting `defconfig (arm64, None)` in the report.  Can we put the build environment string there instead, as that's supposed to be always set to a meaningful value?

mattface: This worked before Christmas, have we stopped parsing the compiler info in the backend?

mattface: That said, I'm happy to use build-environment

gctucker: There's nothing to guarantee the `compiler_version_ext` field will be always in the build meta-data. I ran this locally with build.py and I can see `compiler_version_full` populated in the DB but not the `_ext` one.

gctucker: This patches works for me, using `build_environment` instead: https://termbin.com/fzs6

mattface: Ok, I'm testing locally with build_environment.
As we're requiring that we can guarantee it will be there.

gctucker:
Something should also be done in the boot reports as it currently looks a bit confusing if you have 2 kernels built with different compilers both failing to boot:
```
gtucker/kernelci-local boot: 2 boots: 2 failed, 0 passed (kernelci-local-snapshot-008-10-g2f945ab6e981)

Full Boot Summary: http://192.168.122.1:5000/boot/all/job/gtucker/branch/kernelci-local/kernel/kernelci-local-snapshot-008-10-g2f945ab6e981/
Full Build Summary: http://192.168.122.1:5000/build/gtucker/branch/kernelci-local/kernel/kernelci-local-snapshot-008-10-g2f945ab6e981/

Tree: gtucker
Branch: kernelci-local
Git Describe: kernelci-local-snapshot-008-10-g2f945ab6e981
Git Commit: 2f945ab6e9811142492642568080ea376c16f211
Git URL: https://gitlab.collabora.com/gtucker/linux.git
Tested: 1 unique board, 1 SoC family, 1 build out of 2

Boot Regressions Detected:

arm64:

    defconfig:
        qemu:
            gtucker-lava-testing: new failure (last pass: kernelci-local-snapshot-008-5-ga5f398951cce)

Boot Failures Detected:

arm64:

    defconfig
        qemu: 1 failed lab
```
It says there are 2 failures, but shows only one because the only difference was the compiler used.  So 2 boot tests were submitted and the results were stored correctly in 2 entries, the only issue here is with the report generation.

gctucker:
For the record, I had these 2 builds:
```
gtucker/kernelci-local build: 2 builds: 0 failed, 2 passed (kernelci-local-snapshot-008-10-g2f945ab6e981)

Full Build Summary: http://192.168.122.1:5000/build/gtucker/branch/kernelci-local/kernel/kernelci-local-snapshot-008-10-g2f945ab6e981/

Tree: gtucker
Branch: kernelci-local
Git Describe: kernelci-local-snapshot-008-10-g2f945ab6e981
Git Commit: 2f945ab6e9811142492642568080ea376c16f211
Git URL: https://gitlab.collabora.com/gtucker/linux.git
Built: 1 unique architecture

================================================================================

Detailed per-defconfig build reports:

--------------------------------------------------------------------------------
defconfig (arm64, gcc-bootlin-7.3.0) — PASS, 0 errors, 0 warnings, 0 section mismatches

--------------------------------------------------------------------------------
defconfig (arm64, gcc) — PASS, 0 errors, 0 warnings, 0 section mismatches
```

mattface:
Hopefully this patch should fix the above issue of uniqueness in boot reports. I'd really like to remove HTML reports in the near future because working with the complicated data structures in the backend to produce them is a nightmare.

mattface:
Rebased, PEP8 and unit test compliant. Should be ready for testing again.

app/utils/callback/lava.py
```diff
@@ -74,6 +74,8 @@
     models.VCS_COMMIT_KEY: "git.commit",
     models.IMAGE_TYPE_KEY: "image.type",
     models.PLAN_KEY: "test.plan",
+    models.BUILD_ENVIRONMENT_KEY: "job.build_environment",
+    models.FILE_SERVER_RESOURCE_KEY: "job.file_server_resource"
```
gctucker: It's better to keep a trailing comma to avoid having to edit this line later on when adding extra items to the list.

app/utils/callback/lava.py
```diff
@@ -94,6 +96,8 @@
     models.INITRD_KEY: "job.initrd_url",
     models.BOARD_KEY: "platform.name",
     models.DEVICE_TYPE_KEY: "device.type",
+    models.BUILD_ENVIRONMENT_KEY: "job.build_environment",
+    models.FILE_SERVER_RESOURCE_KEY: "job.file_server_resource"
```
gctucker: Ditto.

app/utils/callback/lava.py
```diff
@@ -222,7 +226,37 @@ def _get_lava_meta(meta, job_data):
...
-def _add_test_log(meta, job_log, base_path, suite):
+def _get_dir_path(meta, base_path):
+    """Parse the meta-data from LAVA
+
+    Update the metadata with the storage path of the artifacts.
+    If possible, use the file_server_resource from the metadata.
+
+    :param meta: The boot meta-data.
+    :type meta: dictionary
+    :param base_path: The .
```
gctucker: The what?

app/utils/callback/lava.py
```diff
@@ -222,7 +226,37 @@ def _get_lava_meta(meta, job_data):
             handler(meta, step["metadata"])
 
 
-def _add_test_log(meta, job_log, base_path, suite):
+def _get_dir_path(meta, base_path):
+    """Parse the meta-data from LAVA
```
gctucker: Looks like this docstring summary was left from a copy-paste and does not reflect what the function `_get_dir_path` actually does.

app/utils/callback/lava.py
```diff
@@ -222,7 +226,37 @@ def _get_lava_meta(meta, job_data):
...
+        dir_path = os.path.join(
+            base_path,
+            meta[models.JOB_KEY],
+            meta[models.GIT_BRANCH_KEY],
+            meta[models.KERNEL_KEY],
+            meta[models.ARCHITECTURE_KEY],
+            meta[models.DEFCONFIG_FULL_KEY],
+            meta[models.BUILD_ENVIRONMENT_KEY],
+            meta[models.LAB_NAME_KEY])
+    meta.update({'dir_path': dir_path})
```
gctucker: This would also work: `meta["dir_path"] = dir_path`.

app/utils/callback/lava.py
```diff
@@ -238,14 +272,7 @@ def _add_test_log(meta, job_log, base_path, suite):
...
 
-    dir_path = os.path.join(
-        base_path,
-        meta[models.JOB_KEY],
-        meta[models.GIT_BRANCH_KEY],
-        meta[models.KERNEL_KEY],
-        meta[models.ARCHITECTURE_KEY],
-        meta[models.DEFCONFIG_FULL_KEY],
-        meta[models.LAB_NAME_KEY])
+    dir_path = meta['dir_path']
```
gctucker: The convention in this code base is to use double quotes everywhere for strings, so `meta["dir_path"]`.

app/utils/callback/lava.py
gctucker: Also, all the other meta-data has its keys defined in `app/models/__init__.py`.  It seems worth adding something like this:
```
STORAGE_PATH = "storage-path"
```
or `DIRECTORY_PATH` or whatever makes sense.

khilman: And please be explicit about any new fields that are mandatory.

mattface: @khilman not sure I understand what you mean

gctucker: There are several types of data fields, but mainly: the JSON payload found in HTTP requests, and the meta-data found in the LAVA job definition.  The former ones have checklists in `app/models/__init__.py` for each request type, and the `BUILD_ENVIRONMENT_KEY` is being added there so that's all fine.  For the latter ones, there isn't any check of this kind so if some mandatory meta-data is missing it will typically cause a Python exception.  This is something that would need to be improved, to return an HTTP error if the data is invalid and document things better.  I don't think there are any new mandatory meta-data fields as part of this PR though.

gctucker:
Tested OK together with #105 and #111 as well as https://github.com/kernelci/kernelci-core-staging/pull/66 on staging.

Here are some builds and boots showing multiple compilers (gcc-7 by default and gcc-8 with some arm64 configs):
https://staging.kernelci.org/build/gtucker/branch/kernelci-local/kernel/kernelci-local-snapshot-025/
https://staging.kernelci.org/boot/all/job/gtucker/branch/kernelci-local/kernel/kernelci-local-snapshot-025/
Note that the boot results appear twice when the same config was built with gcc-7 and gcc-8.  The compiler version is visible in the detailed view.
All pep8 and unit tests are passing at 100%.

PR #87, opened 2018-11-14
T11785 missing test results in report

gctucker:
This is to fix the issue seen with v4l2 test report emails, where some platforms were missing (only Chromebook results were present, not the QEMU ones).  It turned out to be because the number of test groups returned by the mongo query was artificially limited, so some top-level groups were missing.  As there were more test cases when run on QEMU with the `vivid` driver, they were stored later in the database so were not at the top of the list, and as a result were truncated.

### DIFF
--- a/app/utils/report/test.py
+++ b/app/utils/report/test.py
@@ -119,8 +119,6 @@ def create_test_report(data, email_format, db_options,
 
     test_group_docs = list(utils.db.find(
         database[models.TEST_GROUP_COLLECTION],
-        100,
-        0,
         spec=specs,
         fields=TEST_REPORT_FIELDS))
 
--- a/app/utils/db.py
+++ b/app/utils/db.py
@@ -179,7 +179,7 @@ def find_one3(
     return db[collection].find_one(spec_or_id, fields=fields, sort=sort)
 
 
-def find(collection, limit, skip, spec=None, fields=None, sort=None):
+def find(collection, limit=0, skip=0, spec=None, fields=None, sort=None):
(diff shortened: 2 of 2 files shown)


app/utils/report/test.py
```diff
@@ -119,8 +119,6 @@ def create_test_report(data, email_format, db_options,
 
     test_group_docs = list(utils.db.find(
         database[models.TEST_GROUP_COLLECTION],
-        100,
```
MiloCasagrande: Did you try passing zero instead? Can't remember which version of the python driver it's in use, but I think by default the underlying library uses 0 for no limit, same as the `skip` parameter.

gctucker: I did try with a 0 limit and got... 0 results back :)

gctucker: Hrm I tried again on my local set-up and now got unlimited results.  I'll see if it's something that changed in the python driver.

MiloCasagrande: I checked on pymongo docs and code, and that's definitely weird then.

The underlying class (`Cursor`) powering the `find` function sets `limit` to 0 by default as a "no limit set" value. Theoretically passing 0 should return all results...

gctucker: OK I'll verify that again on staging and confirm.  Maybe it was a coincidence when I got 0 results.

gctucker:
Thanks @MiloCasagrande for the review - I've pushed a new version which should make it more obvious and easier to use.

MiloCasagrande:
Looks good, thanks for updating it. Merging it.

PR #90, opened 2018-11-16
Require redis==2.10.6 to avoid incompatible upgrades with celery

gctucker:
When running ansible, the version of redis jumped to 3.0.1, which is
not compatible with celery 3.1.25.  Pin down the version of redis to
avoid this issue.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,7 +2,7 @@ Sphinx==1.4.9
 celery[redis]==3.1.25
 docutils==0.12
 futures==3.0.4
-hiredis
+hiredis==0.2.0
 jinja2
 netaddr==0.7.18
 pycares==1.0.0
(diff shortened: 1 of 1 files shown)


MiloCasagrande:
There is also `hiredis` in the requirements list, isn't that impacted?
If not, just go ahead and merge, looks good.

gctucker:
Good point, the `hiredis` version hasn't caused any issue but if not specified it will also get upgraded and could become a problem in the future.  So I'm pinning it down here as well.

On a side note: I think we should refresh this code base in the near future to use Python3 and some more recent version of these packages.  That should come with probably some redesign of a few things, to be able to separate the test results processing (parsing, regression tracking, bisections, email reports...) with the database storage engine.  Something to discuss a bit more on the mailing list.

MiloCasagrande:
> On a side note: I think we should refresh this code base in the near future to use Python3 and some more recent version of these packages. That should come with probably some redesign of a few things, to be able to separate the test results processing (parsing, regression tracking, bisections, email reports...) with the database storage engine. Something to discuss a bit more on the mailing list.

Yeah, it's definitely something that needs to happen. And since we are at it some more code clean up should be done (especially in the database abstraction layer, need to take a new look at the latest python driver).

PR #93, opened 2018-11-30
Ignore lab name when looking for unique bisections

gctucker:
While each bisection still needs to be associated with a specific lab,
oo not run the exact same bisection twice in more than one lab.  Also
ensure there aren't any duplicate bisections being triggered.


gctucker:
Thanks, typo fixed.

PR #95, opened 2018-12-03
decrease verboseness when a test case is added or updated

ana:
Every time a test case is added or updated, the log at level INFO
outputs a line. This can be very noisy, specially with big tests.
Move these lines to DEBUG level, so they will be only show when
the backend has set the debug mode in /etc/linaro/kernelci-backend.cfg

### DIFF
--- a/app/utils/kci_test/__init__.py
+++ b/app/utils/kci_test/__init__.py
@@ -105,11 +105,11 @@ def save_or_update(doc, spec_map, collection, database, errors):
         doc_id = doc_get(models.ID_KEY)
         doc.id = doc_id
 
-        utils.LOG.info("Updating test document with id '%s'", doc_id)
+        utils.LOG.debug("Updating test document with id '%s'", doc_id)
         ret_val, _ = utils.db.save(database, doc)
     else:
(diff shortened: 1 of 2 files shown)


gctucker:
LGTM - in fact I think it's worth doing this also for sub-groups:
```
diff --git a/app/utils/tests_import.py b/app/utils/tests_import.py
index 2f27f89..36d603f 100644
--- a/app/utils/tests_import.py
+++ b/app/utils/tests_import.py
@@ -424,7 +424,7 @@ def update_test_group_add_sub_group_id(
     ret_val = 200
     errors = {}
 
-    utils.LOG.info(
+    utils.LOG.debug(
         "Updating test group '{}' ({}) with sub-group ID {}".format(
             group_name, str(group_id), sub_group_id))
     database = utils.db.get_db_connection(db_options)
```

ana:
Yup, I forgot that and added it later. It's in the force-pushed commit: https://github.com/kernelci/kernelci-backend/commit/c519f81995dd75985ab3d22c27addf74f5858081

ana:
In any case, the good commit is the one shown now :)

PR #98, opened 2018-12-08
Make sure the boot_regression indexes are created

ana:
This should help to optimise database look ups.

### DIFF
--- a/app/handlers/dbindexes.py
+++ b/app/handlers/dbindexes.py
@@ -29,6 +29,7 @@ def ensure_indexes(database):
     """
     _ensure_job_indexes(database)
     _ensure_boot_indexes(database)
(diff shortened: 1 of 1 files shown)


app/handlers/dbindexes.py
```diff
@@ -261,7 +262,7 @@ def _ensure_stats_indexes(database):
         [(models.CREATED_KEY, pymongo.DESCENDING)], background=True)
 
 
-def _ensure_regressions_indexes(database):
+def _ensure_boot_regressions_indexes(database):
     """Ensure indexes exist on the regression collection.
```
gctucker: The docstring should be updated with "on the boot regressions collection" to match.

PR #102, opened 2019-01-03
T12438 utils/report/test: remove support for HTML test reports

gctucker:
HTML email reports have little added value as most kernel mailing
lists do not accept any HTML content at all but only plain text. Also,
the purpose of the email reports is to notify developers of new
features with a small summary.  Detailed views with all the results
can be made with more advanced graphics on a web dashboard with search
capabilities instead.

For these reasons, drop support for HTML test report emails.


gctucker:
Tested on staging, see https://github.com/kernelci/kernelci-backend/pull/103#issuecomment-451446978

gctucker:
We're not sending any test report in HTML format at the moment (only build and boot reports), and the current HTML test email templates are pretty much broken and badly designed.  So we can either fix HTML email reports properly if anyone thinks the effort can be justified, or drop support for them like in this PR.

I'm in favour of dropping them for the reasons explained in the commit message: most mailing lists don't accept HTML contents, short email reports to notify kernel developers of test failures are more effective and a web-hosted dashboard seems a lot more appropriate for HTML formatting.

Maybe we could have another type of HTML email with a summary of all the test results for a kernel revision or for a device type, but that would be for a different audience (hw vendors probably).

gctucker:
Mailing list discussion: https://groups.io/g/kernelci/topic/html_test_email_reports/28964837

PR #108, opened 2019-01-17
Use database connection directly with test groups

khilman:
A previous patch switched to using an existing database connection in
most functions, but missed this one.

Fixes: 55eb5d5eb995 (Use database connection directly in tests entry point functions)

### DIFF
--- a/app/handlers/__init__.py
+++ b/app/handlers/__init__.py
@@ -1,2 +1,2 @@
-__version__ = "2018.12.2"
+__version__ = "2019.1"
 __versionfull__ = __version__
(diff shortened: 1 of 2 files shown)


khilman:
@gctucker This fixes the database connection errors I reported on IRC yesterday.

Curious why I was the first to notice this.  Are there no kernelCI test jobs using test-groups and sub groups?

gctucker:
Thanks for the fix!  Will give it a spin today on staging and aim to merge it before rolling out into production this week.

All the KernelCI test results should be going through this code, so it's probably something to do with `db_options` not always being the same in your environment while this issue remained silent everywhere else due to good luck.  I'll add some debug code to check how this is working on my local instance before applying your fix to double check that.

gctucker:
@khilman This PR was sent for the `staging` branch, please resend it for the `master` branch instead.

The `staging` branch is used to test pull requests on the staging backend before they get merged.

gctucker:
Tested OK on staging, and cherry-picked the fix on the master branch.

PR #109, opened 2019-01-22
T12645 test plan options v4l2-compliance

gctucker:
Add the possibility to have a test email report template and some options specific to some test plans.  This is to make it possible to show arbitrary information such as a description of the tests being run in the email report, and have more flexibility in the email subject than just the plain test plan name.

Ideally, these options should be set in the YAML test plan definition.  This can't be easily used in the kernelci-backend code at the moment, so until a solution has been found a Python dictionary is used in the test.py source code.

Some ideas for a better approach may be to group the kernelci-core and some of the kernelci-backend features (results parsing, regression tracking, bisections, email reports) or split the YAML configuration into a separate repository which would be shared between kernelci-core and kernelci-backend.

Depends on https://github.com/kernelci/kernelci-core-staging/pull/86

### DIFF
--- a/app/utils/report/templates/plans/v4l2-compliance.txt
+++ b/app/utils/report/templates/plans/v4l2-compliance.txt
@@ -1,0 +1,13 @@
+{% extends 'test.txt' %}
+
+{% block plan_description %}
+V4L2 Compliance on the {{ driver_name }} driver.
+
+This test ran "v4l2-compliance -s" from v4l-utils:
+
+    https://www.linuxtv.org/wiki/index.php/V4l2-utils
+
+See each detailed section in the report below to find out the git URL and
+particular revision that was used to build the test binaries.
+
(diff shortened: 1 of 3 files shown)


gctucker:
Thanks for the reviews.  I'll check with Mauro and other linux-media folks if they like this report format before merging, in case we need to tweak the wording or anything.

gctucker:
Email thread about this: https://www.spinics.net/lists/linux-media/msg144737.html

gctucker:
Tested on staging alongside https://github.com/kernelci/kernelci-core-staging/pull/86.  This was discussed with Hans, the only outstanding issue is that test cases don't appear in the same order in the email report as in the test suite.  The current changes can be merged, but the media tree builds & tests won't be turned on again in production until the test case order has been fixed in a follow-up PR.

PR #110, opened 2019-01-24
T12348 duplicate bisections

gctucker:
This PR does 2 things to avoid reporting the same bisection results more than once:

* Limit the number of bisections run with a same good/bad commit range to 3.
The idea being that only one is enough to report a problem, but 3 is a good number in case some of them fail due to infrastructure errors.

* Do not send bisection reports with the same found git commit and related kernel revision more than once.
Regardless of how a bisection was run, if there has been a previous bisection report with the same git commit found and the same initial kernel regression, then subsequent reports are not sent.

### DIFF
--- a/app/utils/bisect/boot.py
+++ b/app/utils/bisect/boot.py
@@ -311,6 +311,7 @@ def create_boot_bisect(good, bad, db_options):
     database = utils.db.get_db_connection(db_options)
     good_commit, bad_commit = (b[models.GIT_COMMIT_KEY] for b in (good, bad))
     spec = {x: bad[x] for x in [
+        models.LAB_NAME_KEY,
         models.DEVICE_TYPE_KEY,
         models.ARCHITECTURE_KEY,
         models.DEFCONFIG_FULL_KEY,
--- a/app/utils/report/bisect.py
+++ b/app/utils/report/bisect.py
@@ -13,12 +13,15 @@
 
 """Create the bisect email report."""
 
+import hashlib
 import json
+import redis
 import os
 
 import models
 import utils
 import utils.db
+import utils.database.redisdb as redisdb
 import utils.report.common as rcommon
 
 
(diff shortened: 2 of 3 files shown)


gctucker:
With this change, I think we can configure the bisection Jenkins jobs to run LAVA tests at the same priority as the regular ones i.e. "medium" instead of "low" as it is currently configured.  This was mentioned during Monday's meeting and in the minutes:
https://groups.io/g/kernelci/message/248

Does anyone have any issues with that?

gctucker:
Tested on staging with a commit to panic during initramfs initialisation on all device-tree platforms.  Got this boot email summary:
```
gtucker/kernelci-local boot: 37 boots: 32 failed
```
This caused a number of regressions with the same commit range, but only 3 bisections were picked randomly and run.  I got a bisection email report:
```Summary:
  Start:      254c173dd6aa NOISE Special cases aren't special enough...
  Details:    https://staging.kernelci.org/boot/id/5c5b3222cf3a0f3dfd111e56
  Plain log:  http://staging-storage.kernelci.org/gtucker/kernelci-local/kernelci-local-snapshot-026-16-g254c173dd6aa/arm/multi_v7_defconfig/lab-collabora-dev/boot-exynos5800-peach-pi.txt
  HTML log:   http://staging-storage.kernelci.org/gtucker/kernelci-local/kernelci-local-snapshot-026-16-g254c173dd6aa/arm/multi_v7_defconfig/lab-collabora-dev/boot-exynos5800-peach-pi.html
  Result:     f8c7c693ba14 BREAK panic on any device tree platform
```
and the other bisections which were run on other platforms landed on the same commit.  Their reports would have been about the same commit, sent to the same people, but they were filtered out:
```
Feb 07 00:12:15 ip-172-30-0-172 kernelci-celery[13614]: [    INFO/MainThread] Bisection report already sent for kernelci-local-snapshot-026-16-g254c173dd6aa: f8c7c693ba14 BREAK panic on any device tree platform
Feb 07 00:13:57 ip-172-30-0-172 kernelci-celery[13614]: [    INFO/MainThread] Bisection report already sent for kernelci-local-snapshot-026-16-g254c173dd6aa: f8c7c693ba14 BREAK panic on any device tree platform
```

PR #112, opened 2019-01-30
T12812 Include test plan name in test report email unique key

gctucker:
When sending a test report, include the test plan name in the unique
Redis key used to filter duplicate email reports.  Otherwise, sending
reports for more than one test plan for a given kernel would be
filtered out.

Fixes: a4ad0be6f511 ("Include single test plan in test reports")


gctucker:
Tested on staging without the hack to allow multiple email reports to be sent.

PR #116, opened 2019-02-14
Display optionally image.type in the test report

touilkhouloud:
Display the image.type information optionally in the test report email
as long as it's not kernel-ci.


khilman:
FYI for @gctucker this allows us to (re)use the email test reports for kernelci-backend installs that are accepting non-kernelCI data.

gctucker:
There already the rootfs link in the test report.  What kind of values can this "image type" parameter have?

khilman:
It's just a(nother) free-form text field.  We're using it for AGL (yocto) to indicate non-kernelCI jobs.  It can also be used by `test.py` to use an alternate email report template.  We have a separate patch for that, but not sure it's appropriate (yet) for upstream kernelci-backend.

gctucker:
If this is always going to be the same for a given test plan, maybe the test plan name could include the image type?  Then you could have a template for that specific test plan like we already have for `v4l2-compliance`.

Also, rather than adding that at the API level, how about simply adding a meta-data field in the LAVA job definition?  There's already this:
https://github.com/kernelci/kernelci-core/blob/master/templates/base/kernel-ci-base.jinja2#L3
```
metadata:
  image.type: 'kernel-ci'
```
You could replace the hard-coded `kernel-ci` string with a template variable and populate that in `lava-v2-jobs-from-api.py`.  The image type may even be added as an attribute to the test plan definition in `test-configs.yaml`.  A simpler approach is to solve that directly in the jinja2 template files, if you define a block around the `image.type: 'kernel-ci'` line and replace it in your test plan template.  But that's not as nice imo.

khilman:
I think there is some confusion.

First, we are indeed already talking about the `image.type` field that's already part of the metadata, and already stored in the backend db.  Currently, It's value is hard-coded to 'kernel-ci', but we have job templating that sets a different value, and we think that value should show up in the test report.

If this patch displayed that value in the test report unconditionally would that be OK with you?  I assumed it would not be useful information in the default test report since it's always hard-coded to kernel-ci.

app/models/__init__.py
```diff
@@ -758,6 +758,7 @@
             DELAY_KEY,
             EMAIL_FORMAT_KEY,
             GIT_BRANCH_KEY,
+            IMAGE_TYPE_KEY,
```
gctucker: Do you really need to add this key here at the API level if the value is already stored in the meta-data?

gctucker:
Then my first question still stands, it would seem worth considering the approach taken for `v4l2-compliance` test plan variants to add arbitrary information such as the image type format.

app/utils/report/templates/test.txt
```diff
@@ -9,6 +9,9 @@ Test results summary
   Kernel:  {{ kernel }}
   URL:     {{ git_url }}
   Commit:  {{ git_commit }}
+{%- if image_type != 'kernel-ci' %}
```
gctucker: I think using `kernel-ci` as a special value here is a bit fragile, if this ever changes in the base template then it'll break.  But on the other hand, showing this field in every test email report would probably be more confusing than helpful.

khilman:
> Then my first question still stands, it would seem worth considering the approach taken for `v4l2-compliance` test plan variants to add arbitrary information such as the image type format.

Can you point me to the code/templates that do this for v4l2-compliance?  I'm not fully understanding your point.

gctucker:
So this is what the v4l2-compliance email reports use:

https://github.com/kernelci/kernelci-backend/blob/master/app/utils/report/templates/plans/v4l2-compliance.txt

which defines the `plan_description` block which then gets included here:
https://github.com/kernelci/kernelci-backend/blob/master/app/utils/report/templates/test.txt#L5

You can define plan-specific templates and variables here:
https://github.com/kernelci/kernelci-backend/blob/master/app/utils/report/test.py#L60

About the comments on the API changes, if the image type is not used to determine which results to include in the report, in other words if the `image.type` value is always the same for a given test plan name, then I think it shouldn't be "leaked" as a new data field in the HTTP request.

PR #118, opened 2019-03-13
Ensure the measurements display in the email reports

touilkhouloud:
This patch adds a support to the 'test.txt' template to report test case
measurement if there is one.

### DIFF
--- a/app/utils/report/templates/test.txt
+++ b/app/utils/report/templates/test.txt
@@ -38,6 +38,12 @@ Test failures
     * {{ tc.name }}: {{ tc.failure_message }}
       {%- endif %}
     {%- endfor %}
+    {%- for tc in t.test_cases %}
+      {%- if tc.measurements %}
+    * {{ tc.name }}:
+      Measurements: {{ tc.measurements[0]['value'] }} {{ tc.measurements[0]['unit'] }}
+      {%- endif %}
+    {%- endfor %}
     {%- for sg in t.sub_groups %} {# sub_groups #}
       {%- if sg.total["FAIL"] or sg.regressions %} {# sg fail or regressions #}
 
@@ -45,6 +51,12 @@ Test failures
         {%- for tc in sg.test_cases %}
           {%- if 'FAIL' == tc.status %}
       * {{ tc.name }}: {{ tc.failure_message }}
+          {%- endif %}
+        {%- endfor %}
+        {%- for tc in sg.test_cases %}
+          {%- if tc.measurements %}
+      * {{ tc.name }}:
+        Measurements: {{ tc.measurements[0]['value'] }} {{ tc.measurements[0]['unit'] }}
           {%- endif %}
         {%- endfor %}
       {%- endif %} {# sg fail or regressions #}


app/utils/report/test.py
```diff
@@ -47,6 +47,7 @@
     models.JOB_KEY,
     models.KERNEL_KEY,
     models.LAB_NAME_KEY,
+    models.MEASUREMENTS_KEY,
```
gctucker: This lists the fields that should be retrieved from the test groups collection.  The `models.MEASUREMENTS_KEY` is not an attribute of the test group model, so I don't think this is needed.  The measurements are part of the test case model, and the test case entries are added by the `_add_test_group_data` function.

touilkhouloud: oky will remove it and try without

touilkhouloud:
Hello @gctucker  I've forward for you an email test report that contain examples for measurements, is it good for you now ? and I've just applied changes in the 'test.txt' file.

gctucker:
@touilkhouloud If only changing the .txt template works then can you please update this PR with the changes you actually have?  The changes to test.py are still here.

Thanks for the email, although it would be better to make that public and available in this PR so everyone can see it.

One issue I can see is that the test email template is designed to only include tests that fail.  With this change, it tries to show the measurements for every test case.  What is the actual intent here?

touilkhouloud:
The measurements are the information that we are looking for either the test fails or passes, what measurement ensure the success of the test or the failure of it.

touilkhouloud:
Example for a test email report :

 AGL-yocto/gerrit-20645 can: 41 tests, 0 regressions (AGL-gerrit-20645-1)

Test results summary
--------------------

  Test:    can
  Tree:    AGL-yocto
  Branch:  gerrit-20645
  Kernel:  AGL-gerrit-20645-1
  URL:     None
  Commit:  agl-commit


1  | dra7-evm               | arm   |  41 total:  39 PASS   2 FAIL   0 SKIP


Test failures
-------------  
1  | dra7-evm               | arm   |  41 total:  39 PASS   2 FAIL   0 SKIP

  Config:      defconfig+CONFIG_AGL=y
  Compiler:    AGL-yocto
  Lab Name:    clabbe-lab
  Plain log:   http://kernelci.dev.baylibre.com:8082/AGL-yocto/gerrit-20645/AGL-gerrit-20645-1/arm/defconfig+CONFIG_AGL=y/AGL-yocto/clabbe-lab/agl-testplan-dra7-evm.txt
  HTML log:    http://kernelci.dev.baylibre.com:8082/AGL-yocto/gerrit-20645/AGL-gerrit-20645-1/arm/defconfig+CONFIG_AGL=y/AGL-yocto/clabbe-lab/agl-testplan-dra7-evm.html


    test_can_bitrate - 10 tests: 9  PASS, 1 FAIL, 0 SKIP
      * can1_xfer_config_range_start: never passed
      * can0_bitrate_range_end:
        Measurements: 5290998.0 bit/s
      * can1_xfer_tx_range_end:
        Measurements: 1909090.0 bit/s
      * can1_xfer_tx_range_start:
        Measurements: 10791.0 bit/s
      * can1_xfer_config_range_start:
        Measurements: 781.0 bit/s
      * can1_bitrate_range_start:
        Measurements: 781.0 bit/s
      * can0_bitrate_range_start:
        Measurements: 781.0 bit/s   

    test_can_modules - 15 tests: 14  PASS, 1 FAIL, 0 SKIP
      * unload_can_module: never passed

gctucker:
> The measurements are the information that we are looking for either the test fails or passes, what measurement ensure the success of the test or the failure of it.

If you also want to show the results with measurements with passing test cases then you need to redesign the test report email, as currently it won't show the test groups for which all test cases passed.

Only showing the failures and regressions was the result of a discussion on the usefulness of the report, especially as test suites can become long.  Also emails appear to be more about notifying people that something has started breaking rather than providing a complete view of all the results.

So if the measurement is used to determine a pass/fail result, then yes it makes sense to show it alongside the failing test cases.  It would be nice to show the threshold or the logic used to determine why that measurement was considered a failure though, or with regressions it would be useful to show the value of the measurement when it last passed.

khilman:
@touilkhouloud can we close this now, in favor of #123 ?

PR #123, opened 2019-03-29
add a new 'test_template' object in the POST/send API

touilkhouloud:
This patch adds a new 'test_template' object in the POST/send API in
order to diversify the use of the test email report feature. So this
gives the ability to choose which are the test results want to be
reported by choosing the test template.


gctucker:
Can this not already be solved by the existing `plan` parameter?

khilman:
@gctucker that would assume that you always want the same email report template for every test plan.  What we're after here is some flexibility in the email report.  

For example, for a given test plan, you might want to generate an email report focused on failures/regressions (current focus of kernelCI.)   But, if you *also* want to send out a separate email report, to a different audience, focused on the measurements taken by passing tests, you need a different email report.

gctucker:
OK thanks for the context, I understand better where this is coming from now.

My only concern is that the template name is an implementation detail of the backend, and it's the only parameter that is not based on data previously provided from client-side.  But I don't see an easy way to solve this right now, I think the whole system for sending email reports will have to evolve soon.  We keep stretching it for test suites that require many more features than the build and boot reports.

app/utils/report/test.py
```diff
@@ -227,7 +227,18 @@ def create_test_report(data, email_format, db_options,
...
 
-    template = plan_options.get("template", "test.txt")
+    if test_template:
+        path = os.getcwd() + "/utils/report/templates"
+        templates = [t for t in os.listdir( path )]
+        test_template = str(test_template) + ".txt"
+        if test_template in templates:
+            template = plan_options.get("template", test_template)
+        else:
+            utils.LOG.warning("This template does not exist")
```
khilman: In the warning log, please add the filename of the template that it was looking for.

khilman:
@gctucker I agree that having the API user have to know the names of email templates on the backend is not great.  But, I don't see an better solution.  At least this will allow us to do some experimentation with email reporting.

khilman:
@gctucker why was this closed?

gctucker:
Ah, I deleted the `staging` branch because now we have `staging.kernelci.org` instead.  This PR should have been sent against `master` instead.

gctucker:
@khilman @touilkhouloud OK I've recreated the `staging` branch as a copy of today's `master` so we can reopen this PR and continue with the review.  The patches can then be picked onto the actual `master` branch once it's merged.

touilkhouloud:
Hello @gctucker, is everything oky for this patch or need other changes ?

gctucker:
@touilkhouloud Some changes were requested by @khilman.  Also it needs to be rebased before it can be merged, and ideally resubmitted for the `master` branch as this PR is currently for `staging`.

khilman:
@touilkhouloud this patch is still not based on the master branch.

PR #124, opened 2019-04-15
Fix failed test cases order in the report.

mgalka:
This change adds "index" field to the TestCase document. The field is supposed to be used for test case ordering.
For LAVA test results the index field stores the "logged" timestamp, but it can store any other value in case other test backend is used.
This commit adds also sub groups sorting in the e-mail test report. It sorts the test subgroups by the "index" value of the first test case in the subgroup.

The aim of the provided changes is to avoid  e-mail report presenting failed test cases in a different order that they were run by the test backend.

The changes should be enough for LAVA, however they may occur insufficient for test backends that run test groups asynchronously.

Failed test cases report before changes:
```
Test failures
------------- 
1  | qemu                   | arm64 | 117 total:  80 PASS   4 FAIL  33 SKIP

  Config:      defconfig+virtualvideo
  Compiler:    gcc (aarch64-linux-gnu-gcc (Ubuntu/Linaro 7.3.0-27ubuntu1~18.04) 7.3.0)
  Lab Name:    mgalka-lava-local
  Plain log:   http://kernelci//local/mgalka/kernelci-local-snapshot-032-5-gcbdb192f6a76/arm64/defconfig+virtualvideo/gcc/mgalka-lava-local/v4l2-compliance-vivid-qemu.txt
  HTML log:    http://kernelci//local/mgalka/kernelci-local-snapshot-032-5-gcbdb192f6a76/arm64/defconfig+virtualvideo/gcc/mgalka-lava-local/v4l2-compliance-vivid-qemu.html
  Rootfs:      http://storage.kernelci.org/images/rootfs/debian/stretch-v4l2/20190213.0/arm64/rootfs.cpio.gz
  Test Git:    git://linuxtv.org/v4l-utils.git
  Test Commit: aa371c995ec2ad70323db00c47b3132002b060b7
                

    Control-ioctls-Input-0 - 6 tests: 4  PASS, 1 FAIL, 1 SKIP
      * VIDIOC_QUERYCTRL: new failure (last pass: kernelci-local-snapshot-032-4-g0227673db427) 

    Control-ioctls-Input-2 - 6 tests: 4  PASS, 1 FAIL, 1 SKIP
      * VIDIOC_QUERYCTRL: new failure (last pass: kernelci-local-snapshot-032-4-g0227673db427) 

    Control-ioctls-Input-1 - 6 tests: 4  PASS, 1 FAIL, 1 SKIP
      * VIDIOC_QUERYCTRL: new failure (last pass: kernelci-local-snapshot-032-4-g0227673db427) 

    Control-ioctls-Input-3 - 6 tests: 4  PASS, 1 FAIL, 1 SKIP
      * VIDIOC_QUERYCTRL: new failure (last pass: kernelci-local-snapshot-032-4-g0227673db427) 
```

Failed test cases report after changes:
```
Test failures
------------- 
1  | qemu                   | arm64 | 117 total:  80 PASS   4 FAIL  33 SKIP

  Config:      defconfig+virtualvideo
  Compiler:    gcc (aarch64-linux-gnu-gcc (Ubuntu/Linaro 7.3.0-27ubuntu1~18.04) 7.3.0)
  Lab Name:    mgalka-lava-local
  Plain log:   http://kernelci//local/mgalka/kernelci-local-snapshot-032-5-gcbdb192f6a76/arm64/defconfig+virtualvideo/gcc/mgalka-lava-local/v4l2-compliance-vivid-qemu.txt
  HTML log:    http://kernelci//local/mgalka/kernelci-local-snapshot-032-5-gcbdb192f6a76/arm64/defconfig+virtualvideo/gcc/mgalka-lava-local/v4l2-compliance-vivid-qemu.html
  Rootfs:      http://storage.kernelci.org/images/rootfs/debian/stretch-v4l2/20190213.0/arm64/rootfs.cpio.gz
  Test Git:    git://linuxtv.org/v4l-utils.git
  Test Commit: aa371c995ec2ad70323db00c47b3132002b060b7
              

    Control-ioctls-Input-0 - 6 tests: 4  PASS, 1 FAIL, 1 SKIP
      * VIDIOC_QUERYCTRL: new failure (last pass: kernelci-local-snapshot-032-4-g0227673db427)           

    Control-ioctls-Input-1 - 6 tests: 4  PASS, 1 FAIL, 1 SKIP
      * VIDIOC_QUERYCTRL: new failure (last pass: kernelci-local-snapshot-032-4-g0227673db427)     

    Control-ioctls-Input-2 - 6 tests: 4  PASS, 1 FAIL, 1 SKIP
      * VIDIOC_QUERYCTRL: new failure (last pass: kernelci-local-snapshot-032-4-g0227673db427)             

    Control-ioctls-Input-3 - 6 tests: 4  PASS, 1 FAIL, 1 SKIP
      * VIDIOC_QUERYCTRL: new failure (last pass: kernelci-local-snapshot-032-4-g0227673db427) 
```

### DIFF
--- a/app/models/__init__.py
+++ b/app/models/__init__.py
@@ -108,6 +108,7 @@ GTE_KEY = "gte"
 HIERARCHY_KEY = "hierarchy"
 ID_KEY = "_id"
 IMAGE_TYPE_KEY = "image_type"
+INDEX_KEY = "index"
 INITRD_ADDR_KEY = "initrd_addr"
 INITRD_INFO_KEY = "initrd_info"
 INITRD_KEY = "initrd"
@@ -897,6 +898,7 @@ TEST_CASE_VALID_KEYS = {
             CREATED_KEY,
             DEFINITION_URI_KEY,
             KVM_GUEST_KEY,
+            INDEX_KEY,
             MAXIMUM_KEY,
             MEASUREMENTS_KEY,
             METADATA_KEY,
@@ -919,6 +921,7 @@ TEST_CASE_VALID_KEYS = {
         CREATED_KEY,
         DEFINITION_URI_KEY,
         KVM_GUEST_KEY,
+        INDEX_KEY,
         MAXIMUM_KEY,
         MEASUREMENTS_KEY,
         METADATA_KEY,
@@ -939,6 +942,7 @@ TEST_CASE_VALID_KEYS = {
         CREATED_KEY,
         DEFINITION_URI_KEY,
         KVM_GUEST_KEY,
+        INDEX_KEY,
         MAXIMUM_KEY,
         MINIMUM_KEY,
         NAME_KEY,
--- a/app/utils/callback/lava.py
+++ b/app/utils/callback/lava.py
@@ -44,6 +44,7 @@ LAVA_JOB_RESULT = {
 TEST_CASE_MAP = {
     models.NAME_KEY: "name",
     models.STATUS_KEY: "result",
+    models.INDEX_KEY: "logged",
 }
 
 TEST_CASE_NAME_EXTRA = {
@@ -432,7 +433,7 @@ def _add_test_results(group, suite_results, suite_name):
     test_cases = []
     test_sets = {}
 
-    for test in tests:
+    for test in reversed(tests):
         test_case = {
             models.VERSION_KEY: "1.1",
             models.TIME_KEY: "0.0",
@@ -446,6 +447,7 @@ def _add_test_results(group, suite_results, suite_name):
             test_case_list = test_sets.setdefault(test_set_name, [])
         else:
             test_case_list = test_cases
+        test_case[models.INDEX_KEY] = len(test_case_list) + 1
         measurement = test.get("measurement")
         if measurement and measurement != 'None':
             test_case[models.MEASUREMENTS_KEY] = [{
--- a/app/utils/kci_test/__init__.py
+++ b/app/utils/kci_test/__init__.py
@@ -215,6 +215,7 @@ def _update_test_case_doc_from_json(case_doc, test_dict, errors):
 
     case_doc.definition_uri = test_dict.get(models.DEFINITION_URI_KEY, None)
     case_doc.kvm_guest = test_dict.get(models.KVM_GUEST_KEY, None)
+    case_doc.index = test_dict.get(models.INDEX_KEY, None)
     case_doc.maximum = test_dict.get(models.MAXIMUM_KEY, None)
     case_doc.measurements = test_dict.get(models.MEASUREMENTS_KEY, [])
     case_doc.metadata = test_dict.get(models.METADATA_KEY, None)
--- a/app/utils/report/test.py
+++ b/app/utils/report/test.py
@@ -13,9 +13,6 @@
 
 """Create the tests email report."""
 
-import json
-import os
-
 import models
 import utils
 import utils.db
@@ -113,6 +110,8 @@ def _add_test_group_data(group, db, spec, hierarchy=[]):
             if regr:
                 regr_count += 1
         test_cases.append(test_case)
+
+    test_cases.sort(key=lambda tc: tc[models.INDEX_KEY])
 
     sub_groups = []
     for sub_group_id in group[models.SUB_GROUPS_KEY]:
--- a/app/models/test_case.py
+++ b/app/models/test_case.py
@@ -51,6 +51,7 @@ class TestCaseDocument(mbase.BaseDocument):
         self._measurements = []
         self._parameters = {}
 
+        self.index = None
         self.definition_uri = None
         self.kvm_guest = None
         self.maximum = None
@@ -236,6 +237,7 @@ class TestCaseDocument(mbase.BaseDocument):
             models.CREATED_KEY: self.created_on,
             models.DEFINITION_URI_KEY: self.definition_uri,
             models.KVM_GUEST_KEY: self.kvm_guest,
+            models.INDEX_KEY: self.index,
             models.MAXIMUM_KEY: self.maximum,
             models.MEASUREMENTS_KEY: self.measurements,
             models.METADATA_KEY: self.metadata,
--- a/app/models/tests/test_test_case_model.py
+++ b/app/models/tests/test_test_case_model.py
@@ -31,6 +31,7 @@ class TestTestCaseModel(unittest.TestCase):
         test_case.attachments = [{"foo": "bar"}]
         test_case.created_on = "now"
         test_case.definition_uri = "scheme://authority/path"
+        test_case.index = 1
         test_case.kvm_guest = "kvm_guest"
         test_case.maximum = 1
         test_case.measurements = [{"foo": 1}]
@@ -52,6 +53,7 @@ class TestTestCaseModel(unittest.TestCase):
             "attachments": [{"foo": "bar"}],
             "created_on": "now",
             "definition_uri": "scheme://authority/path",
+            "index": 1,
             "kvm_guest": "kvm_guest",
             "maximum": 1,
             "measurements": [{"foo": 1}],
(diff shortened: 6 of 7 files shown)


app/models/__init__.py
```diff
@@ -886,6 +887,14 @@
     ]
 }
 
+
```
gctucker: Why was this block moved before `TEST_CASE_VALID_KEYS`?  Looks like this was done by mistake.

app/models/__init__.py
```diff
@@ -886,6 +887,14 @@
...
+STATISTICS_VALID_KEYS = {
+    "GET": [
+        CREATED_KEY,
+        ID_KEY,
+        VERSION_KEY
+    ]
+}
 TEST_CASE_VALID_KEYS = {
```
gctucker: Missing a blank line here, but again that looks like it was moved by mistake.

app/utils/kci_test/__init__.py
```diff
@@ -215,6 +215,8 @@ def _update_test_case_doc_from_json(case_doc, test_dict, errors):
 
     case_doc.definition_uri = test_dict.get(models.DEFINITION_URI_KEY, None)
     case_doc.kvm_guest = test_dict.get(models.KVM_GUEST_KEY, None)
+    case_doc.index = test_dict.get(models.INDEX_KEY, None)
+    print('MG: {} {}'.format(case_doc.index, models.INDEX_KEY in test_dict))
```
gctucker: Please remove debug prints from the PR.

app/utils/report/test.py
```diff
@@ -114,6 +111,11 @@ def _add_test_group_data(group, db, spec, hierarchy=[]):
                 regr_count += 1
         test_cases.append(test_case)
 
+    try:
+        test_cases.sort(key=lambda tc: tc[models.INDEX_KEY])
+    except KeyError:
+        utils.LOG.warning('Index not found for test cases, leaving unsorted.')
```
gctucker: Is this only to be able to deal with older test cases which don't have an `index` field, or is it expected that new test cases also may not have it?

app/utils/callback/lava.py
```diff
@@ -19,6 +19,8 @@
 import json
 import urllib2
 
+import dateutil
```
gctucker: This standard import should be in the list above, between `codecs` and `errno`.

mgalka: I added it in a separate line as `dateutil` is not a standard library module. Should that be mixed together with the others?

app/utils/report/test.py
mgalka: New test cases should always have the `index` field. This is just for the older TCs which may lack one.

app/utils/kci_test/__init__.py
mgalka: Removed

app/models/__init__.py
mgalka: Corrected

app/models/__init__.py
mgalka: Corrected

app/utils/report/test.py
gctucker: OK, then I think we can simplify the code and remove the `try`-`except` clauses as there shouldn't be email reports generated now with old test data.

app/utils/callback/lava.py
gctucker: I would put it in the list at the top, to only differentiate local modules in this project from the ones installed on the system.  If it's a new requirement then it should probably be added to `requirements.txt` though, but I see that `dateutil` is already used in `lava_log_parser.py`.

gctucker: ... and `python-dateutil` is already listed in the `requirements.txt`.

app/utils/callback/lava.py
```diff
@@ -438,6 +440,11 @@ def _add_test_results(group, suite_results, suite_name):
             models.TIME_KEY: "0.0",
         }
         test_case.update({k: test[v] for k, v in TEST_CASE_MAP.iteritems()})
+        try:
+            test_case[models.INDEX_KEY] = dateutil.parser.parse(test_case[models.INDEX_KEY])
```
gctucker: I think it would be a bit better to have the index follow the test case number i.e. 1, 2, 3... rather than a random integer derived from the timestamp.

mgalka: Yes, I have the fix done already. Testing it on a local env.

app/utils/callback/lava.py
```diff
@@ -432,12 +433,14 @@ def _add_test_results(group, suite_results, suite_name):
     test_cases = []
     test_sets = {}
 
-    for test in tests:
+    for idx, test in enumerate(reversed(tests)):
```
gctucker: One issue with this is that the test case index goes through the whole test suite instead of being within each test set / group.  As a result, if you have 5 groups of 10 tests each, indexes in group 2 will start at 10.  So I think we could drop the `enumerate()` approach and do as I'm suggesting below:
```diff
diff --git a/app/utils/callback/lava.py b/app/utils/callback/lava.py
index a39af1dda5c5..e28fd6c6994a 100644
--- a/app/utils/callback/lava.py
+++ b/app/utils/callback/lava.py
@@ -449,6 +449,7 @@ def _add_test_results(group, suite_results, suite_name):
             test_case_list = test_sets.setdefault(test_set_name, [])
         else:
             test_case_list = test_cases
+        test_case[models.INDEX_KEY] = len(test_case_list)
         measurement = test.get("measurement")
         if measurement and measurement != 'None':
             test_case[models.MEASUREMENTS_KEY] = [{
```
Then I'm not sure if the index should start at 0 or 1, but I guess that's rather subjective.  Maybe starting at 1 is a bit more intuitive, in which case it would be `len(test_case_list) + 1`.

mgalka: Ok, it wasn't entirely clear to me if we want to have an index unique over the test suite or a test group. I'll change it then. I think it'd be better to have it starting from 1

mgalka: Changes applied

gctucker: Thanks, testing that again on staging now.

gctucker:
Tested OK on staging, here's an example list of test cases in the `Control-ioctls-Input-0` test group:
```
> db.test_case.find({test_group_id: ObjectId("5cc341a3cf3a0f3e3e257f79")}, {name: 1, index: 1, status: 1})
{ "_id" : ObjectId("5cc341a3cf3a0f3e3e257f7a"), "status" : "FAIL", "name" : "VIDIOC_QUERY_EXT_CTRL/QUERYMENU", "index" : 1 }
{ "_id" : ObjectId("5cc341a3cf3a0f3e3e257f7b"), "status" : "PASS", "name" : "VIDIOC_QUERYCTRL", "index" : 2 }
{ "_id" : ObjectId("5cc341a3cf3a0f3e3e257f7c"), "status" : "PASS", "name" : "VIDIOC_G/S_CTRL", "index" : 3 }
{ "_id" : ObjectId("5cc341a3cf3a0f3e3e257f7d"), "status" : "FAIL", "name" : "VIDIOC_G/S/TRY_EXT_CTRLS", "index" : 4 }
{ "_id" : ObjectId("5cc341a3cf3a0f3e3e257f7e"), "status" : "PASS", "name" : "VIDIOC_UNSUBSCRIBE_EVENT/DQEVENT", "index" : 5 }
{ "_id" : ObjectId("5cc341a3cf3a0f3e3e257f7f"), "status" : "SKIP", "name" : "VIDIOC_G/S_JPEGCOMP", "index" : 6 }
```

gctucker:
Unit tests and pep8 are also now all passing.

gctucker:
Ah, there's just one small problem with the commit message, it seems it was squashed from several commits.  There should only be one "Signed-off-by" line at the end of the commit.

PR #126, opened 2019-04-24
Skip boot conflict data in reports as it's currently broken

gctucker:
Work around a known bug with boot reports whereby an exception is
raised whenever there is a boot conflict.  This regression was
introduced with support for multiple compilers.  It is causing many
boot emails to not be sent and as a side effects some bisections are
not being triggered.


mattface:
Lets roll this out to get all emails working again, I'm back working on this bug so will rebase and update my PR once it's solved properly

PR #130, opened 2019-05-01
T12593 validate callback metadata

gctucker:
This is to return an HTTP error and not process the LAVA callback data when some metadata keys are not valid, either because they can't be accepted or because mandatory ones are missing.


gctucker:
Tested OK on staging.

PR #135, opened 2019-05-21
add a new 'test_template' object in the POST/send API

touilkhouloud:
This patch adds a new 'test_template' object in the POST/send API in
order to diversify the use of the test email report feature. So this
gives the ability to choose which are the test results want to be
reported by choosing the test template.

### DIFF
--- a/app/handlers/send.py
+++ b/app/handlers/send.py
@@ -112,6 +112,7 @@ class SendHandler(hbase.BaseHandler):
             return response
         report_data = {k: j_get(k) for k in report_keys}
 
+        test_template = j_get(models.TEST_TEMPLATE, None)
         email_format = j_get(models.EMAIL_FORMAT_KEY, None)
         email_format, email_errors = _check_email_format(email_format)
         response.errors = email_errors
@@ -154,6 +155,7 @@ class SendHandler(hbase.BaseHandler):
                 "in_reply_to": j_get(models.IN_REPLY_TO_KEY),
                 "subject": j_get(models.SUBJECT_KEY),
                 "format": email_format,
+                "test_template": test_template,
             }
 
             report_type_or_plan = j_get(models.PLAN_KEY, report_type)
--- a/app/models/__init__.py
+++ b/app/models/__init__.py
@@ -171,6 +171,7 @@ TEST_JOB_PATH_KEY = "test_job_path"
 TEST_JOB_URL_KEY = "test_job_url"
 TEST_GROUP_ID_KEY = "test_group_id"
 TEST_GROUP_NAME_KEY = "test_group_name"
+TEST_TEMPLATE = "test_template"
 TEXT_OFFSET_KEY = "text_offset"
 TIME_KEY = "time"
 TIME_RANGE_KEY = "time_range"
@@ -771,6 +772,7 @@ SEND_VALID_KEYS = {
             SEND_BOOT_REPORT_KEY,
             SEND_BUILD_REPORT_KEY,
             SUBJECT_KEY,
+            TEST_TEMPLATE,
             # Bisection keys
             TYPE_KEY,
             ARCHITECTURE_KEY,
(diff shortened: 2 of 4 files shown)


touilkhouloud:
This pull request replaces this PR:  add a new 'test_template' object in the POST/send API #123

gctucker:
It doesn't seem to be breaking any existing functionality, but it would be good to verify that passing an alternative template file name works.  Maybe this could be done with v4l2-compliance?

touilkhouloud:
Hello @gctucker , sorry I didn't get what you want me to change or to do ...

gctucker:
@touilkhouloud How do you test it works?

touilkhouloud:
@gctucker I test this API using curl to send an email test report with the name of the template and it works.

gctucker:
Great, now we need to be able to test this on staging.kernelci.org.  The v4l2-compliance tests have some special test templates, currently hard-coded in `app/utils/report/test.py`.  It would be good to be able to use the API to specify the template name instead.  Can you do that please?

touilkhouloud:
just one issue, the path in this commit is "/utils/report/templates" to be used to look for the template but to use the v4l2-compliance it is under "/utils/report/templates/plans" so all the templates should be under the same directory and it is going to work for sure.

gctucker:
Well, can you make it work by editing `kernel-arch-complete.sh` and using `plans/v4l2-compliance.txt` or by moving the template outside of the `plans` repository in a backend PR?

If we can't use this API feature or there's no way to test it then it's a bit difficult to justify why it should be merged.

touilkhouloud:
yes sure will move it outside the `plans` repository and add a patch to send email with `curl` in `kernel-arch-complete.sh ` just one more point which email should I use just yours and Kevin's for now ?

gctucker:
You can make the change for the `kernelci` tree, it's only sent to the KernelCI team.  This should be a very simple change, let me know if you have any issues.  Thanks!

touilkhouloud:
Hello @gctucker 
https://github.com/kernelci/kernelci-core/pull/120
https://github.com/kernelci/kernelci-backend/pull/145

I hope I did understand you well, thanks

touilkhouloud:
@gctucker I just modified this patch directly ro remove the template part, so the code won't look for the hard-coded template, is it oky this way ?

touilkhouloud:
Hello @gctucker , is everything okay ?

gctucker:
@touilkhouloud Well there are still some changes needed before this can get merged.  I'll see when I can take another look at it.

The key things are to be able to generate the same report as before, not break pep8 and not break unit tests.  At the moment, the email subject string is different, and there are pep8 and unit test errors.  I  don't think I've actually seen an email report for v4l2-compliance sent using this PR on staging.

khilman:
@gctucker where does one see the pep8 and unit-test warnings/errors?  Are these automated somewhere?

khilman:
@gctucker and can you clarify what you mean by "email subject string is different". I don't think that is the case.

gctucker:
There are no automated tests running anywhere at the moment, but they are run in the standard way:
```
pep8 app
python -m unittest discover app
```

OK I see the change that removed the email subject string has now been updated so it's probably fine from that point of view.

However, as far as I can tell there hasn't been any email report sent from staging using this code.

khilman:
In order to see email reports, this `kernelci-core` change is also needed: https://github.com/kernelci/kernelci-core/pull/122

Is the `staging.kernelci.org` branch for kernelci-core tested along with the staging branch for the backend?

touilkhouloud:
Hello @gctucker , I made this two PRs to test this api and it should work for sure this time.
https://github.com/kernelci/kernelci-backend/pull/150
https://github.com/kernelci/kernelci-core/pull/122
After this we can check the emails form staging.

gctucker:
#153 has been merged instead.

PR #138, opened 2019-05-29
app/handlers/callback: warn when invalid/missing metadata

khilman:
Add error messages into the backend logs when the metadata is invalid
or there are missing mandatory keys.

### DIFF
--- a/app/handlers/callback.py
+++ b/app/handlers/callback.py
@@ -167,10 +167,12 @@ class LavaCallbackHandler(CallbackHandler):
         valid, errors = super(LavaCallbackHandler, self).is_valid_json(
             json_obj, **kwargs)
         if not valid:
+            self.log.error("Invalid JSON {}".format(errors))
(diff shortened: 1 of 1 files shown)


khilman:
@gctucker I'm not sure why the `return False` with the error string is not ending up in the backend logs, but adding this definitely works.

gctucker:
@khilman I think the error string is sent back to the client as part of the HTTP response.  It could probably be fixed for all returned errors, to always be printed in the log rather than just in these 2 specific occurrences.

khilman:
Yeah, it would be better if all those HTTP errors/warnings would show up in the backend logs, because sometimes the clients are jenkins jobs which may or may not be logging those kinds of errors.

khilman:
Add warnings for the rest of the errors, and rebased  onto current master.

gctucker:
LGTM and tested OK on staging.

PR #139, opened 2019-06-19
Use the ATTACHMENTS_KEY

touilkhouloud:
THis patch use the ATTACHMENTS_KEY to collect the reference data from a
lava job which is a link to a file geneated in the test_case.

### DIFF
--- a/app/utils/callback/lava.py
+++ b/app/utils/callback/lava.py
@@ -438,21 +438,24 @@ def _add_test_results(group, suite_results, suite_name):
             models.TIME_KEY: "0.0",
         }
         test_case.update({k: test[v] for k, v in TEST_CASE_MAP.iteritems()})
-        test_meta = test["metadata"]
-        if suite_name == "lava":
-            _parse_lava_test_data(test_case, test_meta)
-        test_set_name = test_meta.get("set")
-        if test_set_name:
-            test_case_list = test_sets.setdefault(test_set_name, [])
-        else:
-            test_case_list = test_cases
-        test_case[models.INDEX_KEY] = len(test_case_list) + 1
         measurement = test.get("measurement")
         if measurement and measurement != 'None':
             test_case[models.MEASUREMENTS_KEY] = [{
                 "value": float(measurement),
                 "unit": test["unit"],
             }]
+        test_meta = test["metadata"]
+        if suite_name == "lava":
+            _parse_lava_test_data(test_case, test_meta)
+        reference = test_meta.get("reference")
+        if reference:
+            test_case[models.ATTACHMENTS_KEY] = [reference]
+        test_set_name = test_meta.get("set")
+        if test_set_name:
+            test_case_list = test_sets.setdefault(test_set_name, [])
+        else:
+            test_case_list = test_cases
+        test_case[models.INDEX_KEY] = len(test_case_list) + 1
         test_case_list.append(test_case)
 
     sub_groups = []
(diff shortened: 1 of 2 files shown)


gctucker:
I thought this was a LAVA v1 only feature?

touilkhouloud:
Hello @gctucker , LAVA give the possibilities to send a link as a reference from a specific test case this way  "lava-test-reference TEST_CASE_ID --result pass|fail|skip|unknown --reference URL" as I found attachments in the backend and not been used so I used it for reference but if you want to implement something only to reference without using attachment will do so

gctucker:
I see, so it sounds like something worth adding but maybe not using the `ATTACHMENT_KEY` as that was inherited from LAVA v1.  Maybe it's OK as long as it's not misleading and it doesn't overlap with how this key was used before.  I'm not familiar with the reference feature so I'll take a closer look.

touilkhouloud:
Oky I will wait then for your response to proceed with the ATTACHMENT_KEY or use something else.

khilman:
@gctucker while "ATTACHMENT" might be the legacy name, the ability to pass a URL is still relevant.  What about we merge this patch as is, to complete the feature.  But then do a rename to "REFERENCE_URL" or similar?

app/utils/callback/lava.py
```diff
@@ -453,6 +453,9 @@ def _add_test_results(group, suite_results, suite_name):
                 "value": float(measurement),
                 "unit": test["unit"],
             }]
+        reference = test.get("reference")
+        if reference and reference != 'None':
```
gctucker: `if reference` is enough, as it will do the same if it's `None` or an empty string.

gctucker:
@khilman OK sorry I didn't realise that there was already an `attachments` attribute in the test case model, so yes it makes sense to complete this now if the reference from LAVA v2 matches what this is intended to be used for.

We probably should also rename it to make it clear that it's a list of URLs and not actual attachment files on the server, but that's a bit more up to debate and not something that's really broken either.

khilman:
@gctucker ping

gctucker:
Sorry, hadn't realised a new version had been pushed.  Checking this on staging now.

app/utils/callback/lava.py
```diff
@@ -453,6 +453,9 @@ def _add_test_results(group, suite_results, suite_name):
                 "value": float(measurement),
                 "unit": test["unit"],
             }]
+        reference = test.get("reference")
```
gctucker: This doesn't work, the `reference` is in the test metadata.

khilman:
@gctucker I think this one is ready now.  It's using your suggested patch.

gctucker:
Yes, tested OK on staging.

Extract from a job definition I ran using a modified "simple" test plan:
```yaml
        run:
          steps:
          - lava-test-case whoami --shell whoami
          - lava-test-reference foo-bar --result pass --reference https://github.com/kernelci/kernelci-backend/pull/139
```

And here's the result stored in the Mongo DB:
```
> db.test_case.findOne(ObjectId("5d309ce6cf3a0f50073db695"))
{
	"_id" : ObjectId("5d309ce6cf3a0f50073db695"),
	"name" : "foo-bar",
	"time" : ISODate("1970-01-01T00:00:00Z"),
	"samples_sum" : null,
	"maximum" : null,
	"vcs_commit" : null,
	"attachments" : [
		"https://github.com/kernelci/kernelci-backend/pull/139"
	],
	"minimum" : null,
	"test_group_name" : null,
	"created_on" : ISODate("2019-07-18T16:23:02.183Z"),
	"index" : 2,
	"kvm_guest" : null,
	"version" : "1.1",
	"definition_uri" : null,
	"measurements" : [ ],
	"test_group_id" : ObjectId("5d309ce6cf3a0f50073db693"),
	"metadata" : null,
	"parameters" : {
		
	},
	"samples_sqr_sum" : null,
	"status" : "PASS",
	"samples" : null
}
```

PR #143, opened 2019-07-03
T15051 Fix suite name fixup in lava callbacks

gctucker:
Fix how the LAVA suite name mangling gets converted to drop the number
prefix.  The problem was when a suite name contained '_' it would also
get split and only the first part of the name would be kept,
i.e. "0_foo_bar" test suite name would become "foo" rather than
"foo_bar".  Using str.partition() fixes this problem.


gctucker:
Tested OK on staging with `igt_drm_kms`.

PR #145, opened 2019-07-11
Make a copy of the v4l2-compliance.txt

touilkhouloud:
This patches adds a copy of v4l2-compliance.txt template in the templates
directory to test the a new 'test_template' object in the POST/send API.


gctucker:
Instead of copying the templates in different locations, it should either be moved or the template parameter should be able to have a relative path in its value i.e. `plans/v4l2-compliance.txt`.

touilkhouloud:
will just use the `plans/v4l2-compliance.txt`

PR #146, opened 2019-07-15
Remove tha hard-coded part of the v4l2-compliance plan

touilkhouloud:
In order to test the new 'test_template' object in the POST/send API,
this patch removes the hard-coded part to use the v4l2-compliance
template.

### DIFF
--- a/app/utils/report/test.py
+++ b/app/utils/report/test.py
@@ -54,24 +54,6 @@ TEST_REPORT_FIELDS = [
     models.VERSION_KEY,
 ]
 
-# FIXME: get this from test-configs.yaml
-TEST_PLAN_OPTIONS = {
-    "v4l2-compliance-uvc": {
-        "subject": "v4l2-compliance on uvcvideo",
-        "template": "plans/v4l2-compliance.txt",
-        "params": {
-            "driver_name": "uvcvideo",
-        },
-    },
-    "v4l2-compliance-vivid": {
-        "subject": "v4l2-compliance on vivid",
-        "template": "plans/v4l2-compliance.txt",
-        "params": {
-            "driver_name": "vivid",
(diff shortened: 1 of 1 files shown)


touilkhouloud:
Helo @gctucker  I don't think that there is a need for this pull request because it's hard-coded only in the staging branch as I shouldn't send pull requests for the staging branch there is nothing to change in the master. 
So only these pull request to test things : https://github.com/kernelci/kernelci-core/pull/122

gctucker:
I'm not following this very well, we should be using this new API parameter and remove the hard-coded table in the backend.  It's not just about testing it on staging but actually using it.  We can discuss this on IRC if you want.

app/utils/report/test.py
```diff
@@ -196,9 +177,8 @@ def create_test_report(data, email_format, db_options,
     tests_total = sum(group["total_tests"] for group in groups)
     regr_total = sum(group["regressions"] for group in groups)
 
-    plan_subject = plan_options.get("subject", plan)
-    subject_str = "{}/{} {}: {} tests, {} regressions ({})".format(
-        job, branch, plan_subject, tests_total, regr_total, kernel)
+    subject_str = "{}/{}: {} tests, {} regressions ({})".format(
```
gctucker: That would make a difference in the email report, I guess the subject string should also be passed via the API.

touilkhouloud: I am a bit lost in the code, I will explain what I got:
the hard-coded table is `TEST_PLAN_OPTIONS` , and then there is ` plan_options = TEST_PLAN_OPTIONS.get(plan, {}` and then `plan_subject = plan_options.get("subject", plan)` I thought that they are all related that's why I removed it this way.

PR #141, opened 2019-06-21
Add new /tests/group API handler

mgalka:
This PR contains changes of /tests/ API handling. It makes use of `kci_test` package to keep lava and non-lava test results consistent.

The POST method handler for `/test/group` endpoint has been heavily redesigned. At the moment it doesn't simply create the test_group document when called. At the moment it accepts `test_group` JSON that can contain sub groups (nested test_group JSONs) and test cases in a form of `test_case` documents.

The `/test[s]` endpoint calls the same code as `/test[s]/group` endpoint at the moment and can be used to post test results.

Handler implementation uses the same mechanisms as lava callback implementation to determine which `build` was used for the test.

The test group JSON may contain `'log'` field. Data from the `log` wield are treated as a text log file and saved as provided by the backend.

The easiest way to test current API is to call curl with appropriate data.

Examples:
- Test group document without sub groups and test cases
```
curl -X POST  http://kernelci:5001/test \
    -H 'Content-Type: application/json' \
    -H 'Authorization: KERNELCI_API_TOKEN' \
    -d @- <<'EOF'
     {
        "sub_groups": [],
        "board": "fake-board",
        "build_environment": "gcc-8",
        "name": "Fake-test-group-with-test-cases",
        "metadata": {},
        "initrd_info": null,
        "git_branch": "fake-branch",
        "git_commit": "feeddeadbeef",
        "version": "1.0",
        "mach": null,
        "board_instance": null,
        "job": "fake-job",
        "file_server_resource": null,
        "initrd": null,
        "device_type": "fake-device-type",
        "arch": "fake-arm64",
        "kernel" : "kernelci-local-snapshot-deadbeef",
        "definition_uri": null,
        "defconfig": "defconfig",
        "defconfig_full": "defconfig",
        "image_type": null,
        "lab_name": "fake-non-lava",
        "index": 999,
        "job_id": null,
        "endian": null,
        "test_cases": [],
        "time": 1.0,
        "log": "Fake LOG\nJust some text\nNo parsing"
    } 
EOF

```

- Test group document without subgroups, with test cases
```
curl -X POST  http://kernelci:5001/test \
    -H 'Content-Type: application/json' \
    -H 'Authorization: KERNELCI_API_TOKEN' \
    -d @- <<'EOF'
     {
        "sub_groups": [],
        "board": "fake-board",
        "build_environment": "gcc-8",
        "name": "Fake-test-group-with-test-cases",
        "metadata": {},
        "initrd_info": null,
[...]

PR #149, opened 2019-07-29
Update test API documentation

mgalka:
This PR provides an update of the test API to reflect the changes in from the PR https://github.com/kernelci/kernelci-backend/pull/141

### DIFF
--- a/doc/schema/1.2/post_test_group.json
+++ b/doc/schema/1.2/post_test_group.json
@@ -1,0 +1,97 @@
+{
+    "$schema": "http://api.kernelci.org/json-schema/1.2/post_test_group.json",
+    "id": "http://api.kernelci.org/json-schema/1.2/post_test_group.json",
+    "title": "test_group",
+    "description": "A test group JSON object",
+    "type": "object",
+    "properties": {
+        "version": {
+            "type": "string",
+            "description": "The version number of this JSON schema",
+            "enum": ["1.0", "1.2"]
+        },
+        "name": {
+            "type": "string",
+            "description": "The name given to this test group"
+        },
+        "lab_name": {
+            "type": "string",
+            "description": "The name of the lab executing this test group"
+        },
+        "time": {
+            "type": "number",
+            "description": "The number of seconds it took to execute the entire test group",
+            "default": -1
+        },
+        "job": {
+            "type": "string",
+            "description": "The name of the job (aka the git tree)"
+        },
+        "kernel": {
+            "type": "string",
+            "description": "The name of the kernel or the git describe value"
+        },
+        "defconfig": {
+            "type": "string",
+            "description": "The name of the defconfig"
+        },
+        "defconfig_full": {
+            "type": "string",
+            "description": "The full name of the defconfig, can also contain config fragments information",
+            "default": "The defconfig value"
+        },
+        "arch": {
+            "type": "string",
+            "description": "The architecture type of this board",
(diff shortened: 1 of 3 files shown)


gctucker:
Updated on staging, here's the generated documentation:
https://staging-api.kernelci.org/collection-test-group.html
https://staging-api.kernelci.org/schema-test-group.html

doc/schema/1.2/post_test_group.json
```diff
@@ -0,0 +1,92 @@
...
+        "test_cases": {
+            "type": "array",
+            "description": "The list of test case objects executed by this test group",
+            "items": {"$ref": "http://api.kernelci.org/json-schema/1.0/test_case_get.json"},
+            "additionalItems": true
+        },
+        "sub_groups": {
+            "type": "array",
+            "description": "The list of sub-group objects included in this group",
+            "items": {"$ref": "http://api.kernelci.org/json-schema/1.0/test_group_get.json"},
```
gctucker: Shouldn't this be `json-schema/1.2` instead?

mgalka:
> What about posting log files?
> 
> Also, I thought some other parts of the API had been removed (i.e. posting a single test case...). Does this still need to be done in the code, with a follow-up change in the related documentation?

I'll add log handling.

TBH, I forgot to remove the old low level API (single test case/group manipulation), that's why it's still in the documentation.

gctucker:
@mgalka OK thanks.  So I guess we can have this merged with the things above fixed, and then have a follow-up PR to remove the old API from the code and from the docs.

doc/schema/1.2/post_test_group.json
mgalka: In general the structure of the json-schema directories is:
```
- schema
  |_ 1.0
     |_ attachment.json
     ...
     |_ get_test_group.json
 |_ 1.1
   |_ ...
 |_ 1.2
   |_...
```

Each version subdirectory contains only the files that were changed in this version of API, therefore test_group_get.json exists only in the version 1.0.
When creating version 1.2 I followed the pattern for version 1.1.

The whole structure looks as if each json is versioned **not** the whole API.

So either should stay as it is or we should change the versioning schema.

gctucker: OK I thought the test group API had changed in version 1.2.  I guess it may all just get removed in the follow-up PR.

gctucker:
> > What about posting log files?
> > Also, I thought some other parts of the API had been removed (i.e. posting a single test case...). Does this still need to be done in the code, with a follow-up change in the related documentation?
> 
> I'll add log handling.

@mgalka Any update on adding this?

> TBH, I forgot to remove the old low level API (single test case/group manipulation), that's why it's still in the documentation.

Any luck with removing the deprecated API?

mgalka:
I added "log" JSON property description.

doc/schema/1.2/post_test_group.json
mgalka: > OK I thought the test group API had changed in version 1.2. I guess it may all just get removed in the follow-up PR.

I double checked it and you're actually right. This shouldn't point to the GET request JSON at first. It should point to version 1.2 of the POST JSON.

gctucker:
Thanks, the HTML version will get generated on staging a bit later today when I start another run.

gctucker:
OK this looks good to get merged now.  I guess there will be another round of updates when the old parts of the API get removed, so we should do another detailed review of the docs at that stage.

PR #150, opened 2019-07-30
Move the v4l2-compliance template

touilkhouloud:
In order to use the new test_template api, this template should be in
the same folders as the other templates(the code check if the requested
template exists or not).


gctucker:
#153 has been merged instead.

PR #152, opened 2019-08-02
Multinode

touilkhouloud:
Support lava multinode jobs.

### DIFF
--- a/app/models/test_group.py
+++ b/app/models/test_group.py
@@ -90,6 +90,7 @@ class TestGroupDocument(modb.BaseDocument):
         self.time = -1
         self.vcs_commit = None
         self.warnings = 0
+        self.lava_multinode = {}
 
     @property
     def collection(self):
@@ -206,6 +207,7 @@ class TestGroupDocument(modb.BaseDocument):
             models.VCS_COMMIT_KEY: self.vcs_commit,
             models.VERSION_KEY: self.version,
             models.WARNINGS_KEY: self.warnings,
+            models.LAVA_MULTINODE_KEY: self.lava_multinode,
         }
 
         if self.id:
--- a/app/handlers/callback.py
+++ b/app/handlers/callback.py
@@ -143,6 +143,7 @@ class LavaCallbackHandler(CallbackHandler):
         super(LavaCallbackHandler, self).__init__(*args, **kwargs)
         self._json_obj = None
         self._job_meta = None
+        self._lava_multinode = None
 
     @property
     def json_obj(self):
@@ -151,6 +152,10 @@ class LavaCallbackHandler(CallbackHandler):
     @property
     def job_meta(self):
         return self._job_meta
+
+    @property
+    def lava_multinode(self):
+        return self._lava_multinode
 
     @staticmethod
     def _valid_keys(method):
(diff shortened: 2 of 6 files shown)


app/models/test_group.py
```diff
@@ -90,6 +90,7 @@ def __init__(self, name, lab_name):
         self.time = -1
         self.vcs_commit = None
         self.warnings = 0
+        self.lava_multinode = {}
```
gctucker: The test groups are LAVA agnostic, so leaking LAVA multinode things here is not going to work.

gctucker (changes requested):

app/models/test_group.py
touilkhouloud: I didn't get what you mean, I tested with a multinode job and a none multinode job and there is no error.

gctucker: I mean that this data model should not assume anything about LAVA.  So if you need to add some extra data to each test group or test case, you can do that but don't make it follow the LAVA multinode design.  It needs to be something that works on its own in a self-contained way.

gctucker:
@touilkhouloud Can you please explain how the multinode LAVA data is meant to be used in KernelCI?

touilkhouloud:
The idea come when we noticed that a multinode job has 2 callbacks that shares the same metadata and the json/logs files will have the same name so one will overwrite the other one so we need some data that differs between the 2 callbacks which is in our case the "role" data for ex one is the host and the other is the target, with this data, @khilman suggested that we may store other multinode data as the group size and others.

gctucker:
Is there any plan to get multinode job results sent to kernelci.org?

I would be interested to see what kind of data would be added to the test results as that is what would be driving the changes in the database models.

touilkhouloud:
We still working on creating templates for multinode job, the use of a multinode job is so big and it differs in our case we are using a host and a target, this is a link for the results, http://kernelci.dev.baylibre.com:8080/test/board/sun8i-h2-plus-orangepi-zero/job/stable/kernel/v5.0.21/
it lacks the attachement data not in data base but you may find it in the data base 8082 
I don't know if this is what you asked for but it may help you I guess

app/models/test_group.py
khilman: So what we need is an abstraction that's not LAVA-centric to describe a group of jobs that are all connected/related.

It appears that the LAVA multinode data has a key called `target_group` which is a unique ID (generated by LAVA) that is shared across all sub-jobs of a multi-node job.

For starters, this would be the minimum thing needed to know that a group of (sub)jobs were all part of the same main (multinode) job.

khilman:
> Is there any plan to get multinode job results sent to kernelci.org?

@gctucker yes, the goal is to contribute test results that require multinode to kernelCI.  For starters, the tests we're working on are power measurements during boot and also during specific tests.

> I would be interested to see what kind of data would be added to the test results as that is what would be driving the changes in the database models.

The kind of data needed in the model is primarily a way to find all the (sub)jobs that were part of a single multinode job. That could simply be a unique ID (LAVA seems to provide this already in the callback data for multi-node jobs as the `target_group` key)  In the case of power measurement, that means the job that controlled the measurement equipment and the job on the DUT that was being measured.   

In addition to some unique ID for all related jobs, I think we'll also need some arbitrary string for human-readable display of the related jobs.  In LAVA multinode terms, we would use the "role" name for that (e.g. "client" or "server" in a network test) or ("measurement host" and "DUT") for a power measurement test.

For starters, those are the minimum requirements that would be needed in the data model.

nuclearcat:

PR #157, opened 2019-09-10
Change all .py files to LGPL v2.1+

mattface:
It was agreed to change the licence of kernelci-backend to LGPL v2.1+ to encourage patches from companies that are not happy with AGPL, this patch should cover the move.

As it's such a large edit, this was done by automated scripts, so there may be small errors in email addresses or some missed files. I suggest we merge anyway and clean up any we spot later on.

Also updates all committers copyright on each file header from the git history.

### DIFF
--- a/COPYING
+++ b/COPYING
@@ -1,0 +1,458 @@
+		  GNU LESSER GENERAL PUBLIC LICENSE
+		       Version 2.1, February 1999
+
+ Copyright (C) 1991, 1999 Free Software Foundation, Inc.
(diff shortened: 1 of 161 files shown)


gctucker:
Tested on staging (i.e. nothing obviously broken).

gctucker:
Having the authors' names helps a bit, and that's consistent with kernelci-core.

gctucker:
So it's all good, thanks!

PR #159, opened 2019-09-23
T16555 Add login test case in LAVA callback handler

gctucker:
When parsing the LAVA callback data, add a "login" test case in the
top-level group based on the result of "auto-login-action" by default.
This is to catch early boot problems as in the legacy boot test plan,
before any test could be run.  Note that the "login" test case result
itself is not LAVA-specific, but only where it got its status from in
this implementation.

As part of this change, stop parsing the rest of the "lava" test
results for the "boot" test plan as this is not useful any more.

### DIFF
--- a/app/utils/callback/lava.py
+++ b/app/utils/callback/lava.py
@@ -429,6 +429,22 @@ def _parse_lava_test_data(test_case, test_meta):
         test_case[models.NAME_KEY] = "-".join(extra_name)
 
 
+def _add_login_case(cases, suite_results, name):
+    tests = yaml.load(suite_results, Loader=yaml.CLoader)
+    tests_by_name = {t['name']: t for t in tests}
+    login = tests_by_name.get(name)
+    if not login:
+        return
+    test_case = {
+        models.VERSION_KEY: "1.1",
+        models.TIME_KEY: "0.0",
+        models.INDEX_KEY: len(cases) + 1,
+        models.NAME_KEY: "login",
+        models.STATUS_KEY: login["result"],
+    }
+    cases.append(test_case)
+
+
 def _add_test_results(group, suite_results, suite_name):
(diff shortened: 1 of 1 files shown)


gctucker:
Tested OK on staging, the `login` test case was added to all the test results.  We should run a test with some boards failing on purpose to check they are caught by this - I'll try that with tomorrow's staging run.

gctucker:
Tested with a patch to panic on odroid-xu3, which was caught by the baseline test:
```
run | platform                     | arch   | lab                 | compiler | defconfig                    | errors
----+------------------------------+--------+---------------------+----------+------------------------------+-------
5   | exynos5422-odroidxu3         | arm    | lab-baylibre-dev    | gcc-8    | multi_v7_defconfig           | 1/1   

  Results:     0 PASS, 1 FAIL, 0 SKIP
  Full config: multi_v7_defconfig
  Compiler:    gcc-8 (arm-linux-gnueabihf-gcc (Debian 8.3.0-2) 8.3.0)
  Plain log:   http://staging-storage.kernelci.org/kernelci/staging.kernelci.org/staging-20190925.2/arm/multi_v7_defconfig/gcc-8/lab-baylibre-dev/baseline-exynos5422-odroidxu3.txt
  HTML log:    http://staging-storage.kernelci.org/kernelci/staging.kernelci.org/staging-20190925.2/arm/multi_v7_defconfig/gcc-8/lab-baylibre-dev/baseline-exynos5422-odroidxu3.html
  Rootfs:      https://storage.kernelci.org/images/rootfs/buildroot/kci-2019.02-8-gd700ebb99e8f/armel/baseline/rootfs.cpio.gz  

  1 tests: 0 PASS, 1 FAIL, 0 SKIP
    * login: new failure (last pass: staging-20190925.1)  
```

PR #160, opened 2019-09-24
Remove the AGPL license file

mattface:
As we've moved the whole repo to LGPL, this license file needs to go.


gctucker:
@mattface Why not replace it with the LGPLv2+1 full text instead?

mattface:
Sorry I should have said, the full LGPL text is in the COPYING file already.

To be honest I don't know what convention expects the filename to be, but COPYING was in kernelci-core so I called it that.

broonie:
On Tue, Sep 24, 2019 at 01:02:56PM -0700, Matt Hart wrote:
> Sorry I should have said, the full LGPL text is in the COPYING file already.
> 
> To be honest I don't know what convention expects the filename to be, but COPYING was in kernelci-core so I called it that.

COPYING is part of the autotools standards for source packages,
IIRC LICENSE.md is something github defined.  I guess if we're
already using COPYING for other things that's a good reason to do
so here.

PR #162, opened 2019-09-25
Use file_server_resource when posting build meta-data

gctucker:
Accept the file_server_resource in build POST requests and use it to
determine the location of the build binaries.  This relies on the
client using the same path when pushing the build binaries and then
sending the meta-data, without having the backend to independently
reconstruct the path (and get it wrong in some cases).

This should also help solving issues such as branch names with a slash
or defconfigs with a path to a config fragment.

### DIFF
--- a/app/utils/build/__init__.py
+++ b/app/utils/build/__init__.py
@@ -524,12 +524,8 @@ def import_single_build(json_obj, db_options, base_path=utils.BASE_PATH):
     git_branch = utils.clean_branch_name(git_branch)
 
     if (utils.valid_name(job) and utils.valid_name(kernel)):
-        # New directory structure:
-        # $job/$branch/$kernel/$arch/$defconfig/$environment
-
-        parent_dir = os.path.join(base_path, job, git_branch, kernel, arch)
-        build_dir = os.path.join(
-            parent_dir, defconfig_full or defconfig, build_environment)
+        file_server_resource = j_get(models.FILE_SERVER_RESOURCE_KEY)
+        build_dir = os.path.join(base_path, file_server_resource)
 
         if os.path.isdir(build_dir):
             try:
--- a/app/handlers/tests/test_build_handler.py
+++ b/app/handlers/tests/test_build_handler.py
@@ -139,7 +139,8 @@ class TestBuildHandler(TestHandlerBase):
                 defconfig="defconfig",
                 arch="arch",
                 git_branch="branch",
-                build_environment="build_environment"
+                build_environment="build_environment",
+                file_server_resource="path/to/build/data"
             )
         )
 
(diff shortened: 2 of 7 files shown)


gctucker:
Tested OK on staging, the build was stored successfully and some test jobs were submitted.  They failed to run for other reasons (ramdisk too big with all the modules...).

gctucker:
Next issue along these lines is the links to logs in the test email report, which also reconstructs wrongly the path to the binaries.  That should be fixed by using the `file_server_resource` in a follow-up PR.

app/utils/build/__init__.py
```diff
@@ -524,12 +524,8 @@ def import_single_build(json_obj, db_options, base_path=utils.BASE_PATH):
     git_branch = utils.clean_branch_name(git_branch)
 
     if (utils.valid_name(job) and utils.valid_name(kernel)):
-        # New directory structure:
-        # $job/$branch/$kernel/$arch/$defconfig/$environment
-
-        parent_dir = os.path.join(base_path, job, git_branch, kernel, arch)
-        build_dir = os.path.join(
-            parent_dir, defconfig_full or defconfig, build_environment)
+        file_server_resource = j_get(models.FILE_SERVER_RESOURCE_KEY)
```
khilman: So this is now a mandatory key so should be checked as such.
If it's not mandatory, it should fall back on the previous behavior.

gctucker: Good point, I've changed that and updated the branch.  It should be mandatory as there's no reason for not using afaict.

gctucker:
Tested OK on staging, the log links are now valid using `file_server_resource`:
```
run | platform              | arch  | lab           | compiler | defconfig                    | results
----+-----------------------+-------+---------------+----------+------------------------------+--------
6   | qemu_arm64-virt-gicv2 | arm64 | lab-collabora | gcc-8    | allmodconfig+...gs/defconfig | 0/1    

  Results:     0 PASS, 1 FAIL, 0 SKIP
  Full config: allmodconfig+KCONFIG_ALLCONFIG=arch/arm64/configs/defconfig
  Compiler:    gcc-8 (aarch64-linux-gnu-gcc (Debian 8.3.0-2) 8.3.0)
  Plain log:   http://staging-storage.kernelci.org/kernelci/staging.kernelci.org/staging-20191021.0/arm64/allmodconfig+KCONFIG_ALLCONFIG=arch-arm64-configs-defconfig/gcc-8/lab-collabora/baseline-qemu_arm64-virt-gicv2.txt
  HTML log:    http://staging-storage.kernelci.org/kernelci/staging.kernelci.org/staging-20191021.0/arm64/allmodconfig+KCONFIG_ALLCONFIG=arch-arm64-configs-defconfig/gcc-8/lab-collabora/baseline-qemu_arm64-virt-gicv2.html
  Rootfs:      https://storage.kernelci.org/images/rootfs/buildroot/kci-2019.02-8-gd700ebb99e8f/arm64/baseline/rootfs.cpio.gz  

  1 tests: 0 PASS, 1 FAIL, 0 SKIP
    * login:
        never passed 
```

gctucker:
Fixed unit tests in latest version, doing one last spin on staging now.

gctucker:
Tested OK on staging.

PR #164, opened 2019-10-01
check if message is a array and join it in one string

alexandrasp:
There someting new on the LAVA log that sends as response an
array as message in cases that happen timed out login,
so to handle with this case we are joining all message
that was splited in an array to one string.

Signed-off-by: Alex P <alexandra.pereira@collabora.com>

### DIFF
--- a/app/utils/lava_log_parser.py
+++ b/app/utils/lava_log_parser.py
@@ -85,6 +85,9 @@ def run(log, boot, txt, html):
         raw_ts = DT_RE.match(dt).groups()[1]
         timestamp = "<span class=\"timestamp\">{}  </span>".format(raw_ts)
 
+        if isinstance(msg, list):
+            msg = ' '.join(msg)  
+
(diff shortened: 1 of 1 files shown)


alexandrasp:
@gctucker I can't ask for reviews :/ or do changes in labels, assignees, etc

app/utils/lava_log_parser.py
```diff
@@ -85,6 +85,9 @@ def run(log, boot, txt, html):
         raw_ts = DT_RE.match(dt).groups()[1]
         timestamp = "<span class=\"timestamp\">{}  </span>".format(raw_ts)
 
+        if(isinstance(msg, list)):
```
gctucker: No need for brackets and missing space after `if`.  It should look like this:
```python
if isinstance(msg, list):
    msg = ' '.join(msg)
```

gctucker:
Tested OK on staging.

PR #169, opened 2019-10-11
Rename /etc/linaro to /etc/kernelci

gctucker:
The kernelci-backend settings files should really be stoed in
/etc/kernelci by default rather than /etc/linaro.


gctucker:
Tested OK on staging.

gctucker:
Fixed a typo in the commit message.

PR #173, opened 2019-11-01
T15206 Fix build emails when no errors and no warnings

gctucker:
Adding support for multiple compilers broke the build reports in
sevaral places, this is a simple fix to avoid exceptions when there
are no errors and no warnings in any of the builds.

This code should really need to be reviewed and rewritten with a
simpler implementation.

Fixes: 472c74146a6b ("Fix build emails blaiming wrong compiler")


gctucker:
Tested OK on staging.

PR #175, opened 2019-11-04
HACK - this is to verify Travis CI is correctly set up

gctucker:
DO NOT MERGE - Temporary PR to cause a unit test failure and check it's caught and correctly.


gctucker:
OK the unit test failure was found as expected.

PR #176, opened 2019-11-04
T17213 Add plan_variant key

gctucker:
This is to add the plan_variant meta-data key and use it in test_group, boot and bisection documents.  The definition of a test plan variant is to know which exact test configuration was used from test-configs.yaml, for example the "baseline" test plan actually uses "baseline_qemu" on QEMU devices.  All the variants of a test plan have their results dealt with in the same way, but have parameters to cover different platforms or combinations within that space.

In particular, this enables bisections to reproduce the test jobs with the same test configuration as used for the original tests that failed and caused a regression.  It's needed to be able to use `kci_test` in the automated bisection jobs (`bisect.jpl` in kernelci-core).

### DIFF
--- a/app/models/__init__.py
+++ b/app/models/__init__.py
@@ -161,6 +161,7 @@ NOT_FIELD_KEY = "nfield"
 PARAMETERS_KEY = "parameters"
 PARENT_ID_KEY = "parent_id"
 PLAN_KEY = "plan"
+PLAN_VARIANT_KEY = "plan_variant"
 PRIVATE_KEY = "private"
 PROPERTIES_KEY = "properties"
 QEMU_COMMAND_KEY = "qemu_command"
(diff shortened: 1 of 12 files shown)


gctucker:
Tested OK on staging, here's a sample of the test groups in the backend database:
```
> db.test_group.find({kernel: 'staging-20191105.2'}, {name: 1, plan_variant: 1})
{ "_id" : ObjectId("5dc14419cf3a0f3cea9243dd"), "plan_variant" : "boot", "name" : "boot" }
{ "_id" : ObjectId("5dc144bacf3a0f3cea9243df"), "plan_variant" : "baseline", "name" : "baseline" }
{ "_id" : ObjectId("5dc14677cf3a0f3cea9243e2"), "plan_variant" : "baseline_qemu", "name" : "baseline" }
{ "_id" : ObjectId("5dc1469bcf3a0f3ce99243df"), "plan_variant" : "boot_qemu", "name" : "boot" }
```

gctucker:
Getting this tested on staging with a bisection.

gctucker:
Tested OK on staging.

PR #178, opened 2019-11-22
PyMongo upgrade

gctucker:
Upgrade to PyMongo v3.9 which is requied to use the latest version of Mongo DB (v4.2) which is in turn required when running on Debian Buster in order to get updates and keep the backend server going.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,11 +15,8 @@ script:
   - >
     for dir in
     models/tests
-    utils/tests
-    utils/build/tests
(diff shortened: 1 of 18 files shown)


gctucker:
Tested OK on staging with all the main features working.

Things that still need to be fixed;
* unit tests with newer mongomock
* list of compilers doesn't show up on frontend
* pie chart with build results doesn't work, it always shows 0 / 0 / 0

PR #179, opened 2019-11-25
Fix string formating with missing build_environment

gctucker:
Add missing build_environment to string formatting in error message.

Fixes: cb89ab48861b ("Add a build environment required field to builds")


gctucker:
Tested OK on staging.

PR #182, opened 2019-11-27
Use yaml.safe_load() in utils.report.test

gctucker:
Fix this warning by using safe_load instead:
```
YAMLLoadWarning: calling yaml.load() without Loader=... is deprecated
```


gctucker:
Tested OK on staging.

PR #183, opened 2019-12-02
Remove redundant key when updating test group document IDs

gctucker:
The "kernel" key is already present in the initial spec to find the
job document so there's no need to update it again to search for a
build document.


gctucker:
Tested OK on staging.

PR #184, opened 2019-12-03
app/handlers/callback: use yaml.safe_load

khilman:
When testing a fresh docker-based install using debian/buster and
python3, this YAML load triggers an error about the missing CLoader
as well as the unsafe load.  Fix it.


khilman:
False alarm.  CLoader issue triggered by missing libyaml-dev package when python dependencies were installed.

PR #185, opened 2019-12-03
python3 rework: convert to except as

khilman:
"except Foo, ex" is a syntax error in python3.
Convert all occurances to "except Foo as ex"


khilman:
Tested with python2.7 also.

khilman:
Not tested enough, needs more work.

PR #187, opened 2019-12-06
callback/lava: use board-instance for log files

khilman:
Currently the log files are generated based on the LAVA device-type
(models.BOARD_KEY, which comes from the platform.name metadata.)

If there are multiple instances of a given device-type in a lab, the
log files will be identical, so will clash and the most recently
arrived job will overwrite any logs from previous jobs.

To avoid this, use BOARD_INSTANCE_KEY instead of BOARD_KEY when
generating the log file names.

While here, ensure that the BOARD_INSTANCE_KEY is among the sub_group
keys updated when new test reports arrive.

### DIFF
--- a/app/utils/callback/lava.py
+++ b/app/utils/callback/lava.py
@@ -292,7 +292,7 @@ def _add_test_log(meta, job_log, suite):
     dir_path = meta[models.DIRECTORY_PATH]
 
     utils.LOG.info("Generating {} log files in {}".format(suite, dir_path))
-    file_name = "-".join([suite, meta[models.BOARD_KEY]])
+    file_name = "-".join([suite, meta[models.BOARD_INSTANCE_KEY]])
     files = tuple(".".join([file_name, ext]) for ext in ["txt", "html"])
     meta[models.BOOT_LOG_KEY], meta[models.BOOT_LOG_HTML_KEY] = files
     txt_path, html_path = (os.path.join(dir_path, f) for f in files)
@@ -322,7 +322,7 @@ def _store_lava_json(job_data, meta, base_path=utils.BASE_PATH):
     :type base_path: string
     """
 
-    file_name = "-".join(["lava-json", meta[models.BOARD_KEY]])
+    file_name = "-".join(["lava-json", meta[models.BOARD_INSTANCE_KEY]])
     file_name = ".".join([file_name, "json"])
 
     dir_path = meta[models.DIRECTORY_PATH]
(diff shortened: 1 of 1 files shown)


khilman:
I'm not aware of other tooling relying on the format of these logfile names, but there maybe something I'm forgetting about.

gctucker:
Tests are currently run on a per-device-type basis, the fact that there are several instances of a same device type in a lab is not tracked by KernelCI.  Each test is run only once on a given device type.  Also, adding this here and tracking each individual device instance would make the rest of the database schema and reporting out-of-sync as it would be a massive shift compared to what we've been doing so far.  Running the tests on all the available devices in each lab would also increase the load significantly and probably not add much value.

Is there a practical use-case that made you think that logs were being over-written?

khilman:
There is no existing use-case in KernelCI today where the logs are overwritten.  However, it's a very real potential when using the backend with job schedulers other than the official, upstream kernelci-backend which occured to me when working on the kcidb code for handling logs.

I think pre-preemptively ensuring that this cannot happen (especially with such a trivial patch) should be valuable, even if we don't have the problem today (in upstream KernelCI.)  

IMO, In general, the backend should not hard-code assumptions about how jobs are currently scheduled in labs.

khilman:
Also, since the log files are currently not used by any automated tooling, now is a good time to fix these kinds of issues.

gctucker:
It's all about defining what we want to do.  At the moment, we only consider device types but not individual boards.  The log files are just one aspect of it, test results are also only stored on a per-device-type basis.

If you run the same test in the same lab with different device instances, you would still only get one set of results in the database.  The second run would overwrite the first one, because that's the current convention.

If we moved the convention that tests are on a per-device-instance basis, then running the same test twice on different instances would indeed be stored separately but a lot of other things would also need to be updated too in order to make use of that.

While this is a very small change, I wouldn't say it's a trivial one because it goes against the current convention of running tests on a per-device-type basis.

khilman:
I agree that it's not a priority to change the way we store this in the database, but IMO, that is a separate topic.  The goal of this PR is just to have (more) clarity in the log file names.

But, if you think we can't change this unless we change everything at the same time, I'll survive, though I don't agree with that reasoning.

gctucker:
I think we can change the file names in practice to include the board instance name without necessarily changing anything else.  But again, I fail to see how it would help in any way if we don't also change how the results are dealt with to be on a per-board basis.

With all that being said, if you still feel like the log file should really include the board instance name, we can still merge this PR.  It seems to me more like a risk to break something by accident, or confuse things by using the board instance name in the log while the results use the device type, when I don't really understand what net benefit we would get by doing it.  As there doesn't seem to either be any absolute technical reasons for it nor against it, it's also a somewhat subjective point of view.

khilman:
OK, agreed.  Subjective and not urgent.

PR #189, opened 2019-12-19
Fix unittests

mgalka:
This PR adds more unittest fixes for kernelci-backend

Fix mock to patch correct DB connecting function. 

New PyMongo API requires filter parameter when counting
documents. Added empty dictionary to satisfy this requirement.

There is a bug in mongomock implemenation that returns negative value
when number of skipped documents is bigger that the number of documents
found. Skipping the test that uses it till it's fixed in Mongomock.

**mongomock bugged code excerpt**

```python
count = len(list(self._iter_documents(filter)))
        if limit is None:
            return count - skip
        return min(count - skip, limit)
```

### DIFF
--- a/app/handlers/tests/test_build_handler.py
+++ b/app/handlers/tests/test_build_handler.py
@@ -152,7 +152,7 @@ class TestBuildHandler(TestHandlerBase):
             response.headers["Content-Type"], self.content_type)
 
     def test_delete(self):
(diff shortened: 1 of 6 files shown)


gctucker:
Can we now enable Travis CI unit tests in `app/handlers/tests`, or are there still other broken unit tests there?

PR #192, opened 2019-12-30
Enable handler/tests in TravisCI

mgalka:
Re-enables handler/tests in TravisCI and sets six package version
to 1.13.0. Versiion 1.11.0 of the six package was implicitly causing  installed
it's now set explicitly to 1.13.0 in requirements.txt.


gctucker:
Tested OK on staging.

PR #197, opened 2020-01-10
T18078 utils.callback.lava: remove unused _parse_lava_test_data() function

gctucker:
Remove the utils.callback.lava._parse_lava_test_data() function as
it's not called and not needed any more.  Only the login test cases
uses the LAVA test data to populate test results.


gctucker:
Tested OK on staging, with all test results handled correctly as well as a `baseline.login` bisection.

gctucker:
Tested OK again on staging.

PR #198, opened 2020-01-14
T18021 remove deprecation warning

mgalka:
### Switch from deprecated pymongo methods to new PyMongo API
    
Switched from collection.remove to collection.delete_many. The change of
the PyMongo API required also modification of lab and token DELETE
handlers.

### Remove DELETE handlers for build, job and boot documents.
    
There are no use cases for DELETING these documents, apart from that
deletion of boot and job documents with the current handlers might have
lead to data integrity issues. Surplus tests are removed, too.


gctucker:
Tested OK on staging.

app/handlers/lab.py
```diff
@@ -150,17 +150,17 @@ def execute_delete(self, *args, **kwargs):
 
         return response
 
-    def _delete(self, spec_or_id, **kwargs):
+    def _delete(self, id_, **kwargs):
```
gctucker: Maybe call it `oid` rather than `id_`?

app/utils/db.py
```diff
@@ -503,23 +504,34 @@ def find_and_update(collection, query, document, operation="$set"):
     return ret_val
 
 
-def delete(collection, spec_or_id):
-    """Remove a document or multiple documents from the collection.
+def delete_by_id(collection, id_):
```
gctucker: Ditto - maybe `oid` would be better than `id_`.

gctucker:
Tested OK on staging.

app/handlers/lab.py
mgalka: `oid` makes more sense. Corrected.

app/utils/db.py
mgalka: Corrected.

gctucker:
Tested OK again on staging.

PR #199, opened 2020-01-16
T18078 test case clean up

gctucker:
Remove unused fields from TestCaseDocument.  These have never been used in production.


gctucker:
Tested OK on staging.

PR #200, opened 2020-01-21
T18187 test case counts

gctucker:
Add fields to TestCaseDocument to be able to count and filter all test cases for a given kernel revision or test plan.  Also allow queries to count test regressions and use "null" to specify fields that need to be None.


gctucker:
Tested OK on staging.

PR #201, opened 2020-01-22
T18187 Stop storing boot results in test documents

gctucker:
There was a plan to keep boot tests but treat them as a special test
plan.  This is now being changed to use the baseline test plan
instead, with the special "login" test case, and to remove boot tests.

As such, stop storing boot results in test case documents as it is not
useful any more.


gctucker:
Tested OK on staging.

PR #202, opened 2020-01-22
Fix collection name in index creation handler

gctucker:
The collection name was hard-coded to "builds" due to a mistake in a
previous commit.  Fix the collection name used when creating indexes
to create all the required indexes as intended.

Fixes: 6fd607fdbee3 ("Use Collection.create_index() rather than .ensure_index()")


gctucker:
Note: this didn't have much effect on performance in production as the indexes already existed in the database before the bug was introduced.  This fix will have a big impact on new databases, or new indexes in existing ones.

gctucker:
Tested OK on staging, the new indexes from #200 were created as expected.

PR #204, opened 2020-01-23
T18689 test case plan

gctucker:
Add a "plan" field to TestCaseDocument to be able to filter test cases using a test plan name.


gctucker:
Sample test case document:
```
> db.test_case.findOne({kernel: 'staging-20200123.1'})
{
	"_id" : ObjectId("5e2a134783dd95b2b9664d8b"),
	"version" : "1.0",
	"job" : "kernelci",
	"test_case_path" : "baseline.login",
	"test_group_id" : ObjectId("5e2a134783dd95b2b9664d8a"),
	"created_on" : ISODate("2020-01-23T21:42:31.438Z"),
	"time" : ISODate("1970-01-01T00:00:00Z"),
	"measurements" : [ ],
	"plan" : "baseline",
	"git_branch" : "staging.kernelci.org",
	"regression_id" : null,
	"status" : "PASS",
	"index" : 1,
	"kernel" : "staging-20200123.1",
	"name" : "login"
}
```

For the next step about showing more details about test results, we might need to also add things that are only currently available in top-level test group documents such as the `device_type`, `defconfig`, `build_environment`, `lab`.  It would probably be more efficient to get this information from the top-level group document if possible though, so I think it's worth trying that first.

PR #205, opened 2020-01-24
WIP T18187 test case soc

gctucker:
These changes are required to handle per-soc test cases.  This is especially useful for showing the test results summaries on the SoC view on the frontend.


gctucker:
Replaced with #206

PR #206, opened 2020-01-24
T18732 test case mach

gctucker:
These changes are required to handle per-soc test cases. This is especially useful for showing the test results summaries on the SoC view on the frontend.


gctucker:
> batmobile, raspberry and laptop? Maybe you should unify all of the defaults?

These are random values used in unit tests.

PR #208, opened 2020-01-28
T18074 test plan indexes

gctucker:
Add Mongo DB indexes and enable test group fields to be able to create views based on test plans.


gctucker:
Tested OK on staging with https://github.com/kernelci/kernelci-frontend/pull/63.

PR #210, opened 2020-01-31
T18879 Drop support for compare queries

gctucker:
The "compare" feature is not currently working, and it has never
actually been fully enabled.  Since this is effectively a lot of
unused code, drop the feature altogether until some new plans are made
for a better web frontend with better defined use-cases.


gctucker:
Tested OK on staging.

PR #211, opened 2020-02-07
T18935 Add test/regression API entry point

gctucker:
Add API entry point to retrieve test regressions and also enable batch
operations on them.


gctucker:
Tested OK on staging.

PR #213, opened 2020-02-13
T18935 Decode query values in batch requests

gctucker:
When processing queries in batch requests, decode the values as if
they had been passed in a URL.  This is to allow special characters
such as '&' and '=' to be used in the query values without breaking
the parsing.  They need to be encoded as per RFC 3986.

Also simplify the logic for constructing a dictionary of lists.


gctucker:
Tested OK on staging.

PR #215, opened 2020-02-14
Bump Mongomock to 3.19.0

mgalka:
This version contains fix for documenents_count(), so we can re-enable
the test that was skipped.


gctucker:
LGTM - thanks!

PR #216, opened 2020-02-18
Add line numbers to the HTML log file

touilkhouloud:
To be able to point to a specific line in the html log file, we need to
add the line numbers which could be referenced later in the URL.

### DIFF
--- a/app/utils/lava_log_parser.py
+++ b/app/utils/lava_log_parser.py
@@ -48,6 +48,7 @@ HTML_HEAD = """\
   a:link {{text-decoration: none }}
   a:visited {{text-decoration: none }}
   a:active {{text-decoration: none }}
+  :target {{ background-color: #000080; }}
   a:hover {{text-decoration: bold; color: red; }}
   ul.results {{
     list-style-type: none;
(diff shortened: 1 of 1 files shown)


app/utils/lava_log_parser.py
```diff
@@ -122,27 +124,33 @@ def run(log, boot, txt, html):
...
     start_ts = None
     log_buffer = []
+    lineno = 0
 
     for line in log:
+        lineno = lineno + 1
         dt, level, msg = (line.get(k) for k in ["dt", "lvl", "msg"])
         raw_ts = DT_RE.match(dt).groups()[1]
         timestamp = "<span class=\"timestamp\">{}  </span>".format(raw_ts)
+        sl = "<p id=%d>" % lineno
```
gctucker: Maybe use another type of HTML element instead of `<p>` with zero margin.  Paragraphs are supposed to have space between them.  The log could be a list and each line an item, or you could use `<span>` which doesn't add any formatting.

gctucker: Also, using just the plain number as the `id` value may not be very wise as these are global identifiers for the document.  Rather than `id=123`, maybe do like LAVA does with `id="L123"` or something even more explicit such as `id="line-123"`.

app/utils/lava_log_parser.py
```diff
@@ -122,27 +124,33 @@ def run(log, boot, txt, html):
...
+        el = "</p\n>"
 
         if isinstance(msg, list):
             msg = ' '.join(msg)
 
         fmt = formats.get(level)
         if fmt:
-            log_buffer.append(timestamp + fmt.format(cgi.escape(msg)))
+            log_buffer.append(
```
gctucker: I think that with that number of elements, using `join()` is a bit more efficient:
```python
''.join([sl, timestamp, fmt.format(cgi.escape(msg)), el])
```

gctucker:
Tested OK on staging.  Here's an example:
https://storage.staging.kernelci.org/kernelci/staging.kernelci.org/staging-20200225.0/arm/multi_v7_defconfig/gcc-8/lab-collabora/igt-kms-tegra-tegra124-nyan-big.html#L123

PR #219, opened 2020-02-25
kcidb: add valid, waived fields

khilman:
The WIP Grafana frontend for kcidb is using the "valid" field for
revisions and "waived" field for tests for filtering.  Set them
appropriately.


gctucker:
Tested OK on staging.

PR #223, opened 2020-03-12
Add the test runner url to the email report

touilkhouloud:
To be able to add the test runner url (ex: Jenkins url) to the email
report, we had to add it first to the test group data base and then
use the field in the email report

### DIFF
--- a/app/models/__init__.py
+++ b/app/models/__init__.py
@@ -180,6 +180,7 @@ TEST_JOB_ID_KEY = "test_job_id"
 TEST_JOB_PATH_KEY = "test_job_path"
 TEST_JOB_URL_KEY = "test_job_url"
 TEST_GROUP_ID_KEY = "test_group_id"
+TEST_RUNNER_URL_KEY = "test_runner_url"
 TEXT_OFFSET_KEY = "text_offset"
 TIME_KEY = "time"
 TIME_RANGE_KEY = "time_range"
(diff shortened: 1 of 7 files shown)


khilman (approved):
LGTM.  I'm not sure if "test_runner_url" is the best name, but it's better and more generic than "jenkins_url" since jenkins is not a requirement.  

@gctucker any opinions on the name here?

touilkhouloud:
Hello, @gctucker please can you have a look on this PR ?

khilman:
@gctucker ping for opinions on naming.

gctucker:
To be honest, I don't understand what this is all about.  What do you call a test runner?  I don't see how a Jenkins URL could be useful to anyone except KernelCI admins so I think I'm missing some context here.

khilman:
It's useful for KernelCI admins, but also anyone working on their own instance of KernelCI and wanting to be able to work backwards in the pipeline from a failed build to the original "test runner" (e.g. Jenkins job.)

For example, for AGL, we would like to do a modified email report templated, and they would like this URL in the email reports.

This is particularily useful for those just being introduced to KernelCI.  They may be already familiar with their jenkins pipeline, so having a link back to Jenkins will greatly help developer debug.

gctucker:
OK, I see.  "Test runner" sounds be a bit ambiguous though, as it can easily be mixed up with a device running the tests such as board in a LAVA lab or a VM somewhere.  Maybe orchestration would be a better word.

We shouldn't have this link appearing to general email reports, and I see it is optional so that should be fine.  It would be good to have at least one test run with this URL set on staging to test it actually works end-to-end once.

nuclearcat:

PR #226, opened 2020-03-18
Add test_group.modules and .modules_dir fields

gctucker:
Add .modules and .modules_dir fields to TestGroupDocument and populate their values from the associated build document to be able to tell which modules were used in tests.  Update unit tests accordingly.


gctucker:
This is actually not necessary, it's easy enough to get it from the group document.

The whole test case / test group schema should be revisited once all the frontend work is complete, keeping backend changes to only what is really needed for now.

PR #230, opened 2020-04-07
Add Mongo DB index in test cases with job/branch/date/status for trends

gctucker:
Add an index for the job/branch/date/status values in test case
entries which is used typically for creating graphs with pass/fail
trends over results sorted by date.  Without this index, the following
error could be seen:

    OperationFailure: Sort operation used more than the maximum
    33554432 bytes of RAM. Add an index, or specify a smaller limit.


gctucker:
The error quoted in the commit message is caused by this view:
https://staging.kernelci.org/job/kernelci/branch/staging.kernelci.org/

gctucker:
Tested OK on staging.  The errors are gone and the trends graph for test results pass/fail rate is loading.

PR #232, opened 2020-04-15
T18940 sub group fields

gctucker:
Add missing group and sub-group fields to show all the relevant information directly on the frontend without having to do extra queries.


app/models/tests/test_test_group_model.py
```diff
@@ -111,6 +112,7 @@ def test_group_doc_to_dict(self):
             "kernel_image": "kernel-image",
             "lab_name": "lab-name",
             "mach": "mach",
+            "modules": "moduli",
```
alexandrasp, line 115: @gctucker any special reason to use this typo in test?

gctucker: Yes, to use an arbitrary value that is not likely to be used anywhere else otherwise.  This is to avoid having default values being set and hiding an issue.  It's also the plural of `modulus` :)

PR #233, opened 2020-04-21
Update kcidb-submit options for kcidb v5

gctucker:
With the kcidb v5 release, the kcidb-submit command line options have
changed.  Instead of having a BigQuery dataset name, there is now a
project "-p" and a topic "-t".  These are kcidb concepts rather than
BigQuery, so rename the celery config options to kcidb_options and use
project/topic instead of dataset.  The credentials is still an actual
BigQuery credentials so that doesn't change.

### DIFF
--- a/app/taskqueue/tasks/kcidb.py
+++ b/app/taskqueue/tasks/kcidb.py
@@ -25,16 +25,17 @@ import utils.kcidb
 @taskc.app.task(name="kcidb-build")
 def push_build(args):
     build_id, job_id, first = args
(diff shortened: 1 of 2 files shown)


gctucker:
Tested OK on staging, no errors in the kcidb Celery task and results are available on the Grafana dashboard: https://staging.kernelci.org:3000/d/revision/revision?orgId=1&var-id=kernelci:staging.kernelci.org:v5.7-rc2-79-gc578ddb39e56

PR #237, opened 2020-05-04
T21027 Add serial batch request handler

gctucker:
Rather than setting up a group of Celery tasks to process the batch
requests in parallel, also provide a single Celery task to process
them in series.  This turns out to be a lot faster when the processing
that needs to be done is relatively small due to the big overhead of
setting up a parallel group and putting the results together.

- [ ] timing compared on staging with parallel handler
- [ ] unit tests updated


gctucker:
There is no clear evidence that batch requests are slower than they should and that removing the overhead of setting up parallel tasks is actually improving things at all with a real workload.

PR #242, opened 2020-05-18
T20795 order by field on aggregate

alexandrasp:
this commit updates aggregate asking for $max or $min in a group in order to always have  the expected value considering that group doesn't maintain the order.

Signed-off-by: Alexandra Pereira <alexandra.pereira@collabora.com>

### DIFF
--- a/app/utils/db.py
+++ b/app/utils/db.py
@@ -601,6 +601,14 @@ def aggregate(
         else:
             return _starts_with_dollar(unique)
 
+    def _parse_sort_field(group_dict, sort):
+        for item in sort:
+            field, sort_order = item
+            if sort_order > 0:
+                group_dict['$group'][field] = {'$min': '$'+field}
+            else:
+                group_dict['$group'][field] = {'$max': '$'+field}
+
     # Where the aggregate actions and values will be stored.
     # XXX: The append order is important!
     pipeline = []
@@ -630,6 +638,10 @@ def aggregate(
             ]
 
     group_dict["$group"].update(r_fields)
+
+    if sort:
+        _parse_sort_field(group_dict, sort)
+
     pipeline.append(group_dict)
 
     # Sort everything now.


alexandrasp:
@gctucker seems that everything works fine on staging, for all soc branches list we have latest test result with right date. Not sure about the performance..but since `$max` and `$min` are attributes for aggregation on mongodb, seems to be ok.

gctucker:
Yes sounds good, thanks.  I'll take a closer look to see if it causes any side effects and leave it running on staging for a day or two.

alexandrasp:
absolutely and checking if all on page is right.

app/utils/db.py
```diff
@@ -630,6 +638,7 @@ def _parse_aggregate():
             ]
 
     group_dict["$group"].update(r_fields)
+    _parse_sort_field(group_dict, sort)
```
gctucker, line 641: Add `if sort:` as it could be `None`.

app/utils/db.py
```diff
@@ -601,6 +601,14 @@ def _parse_aggregate():
         else:
             return _starts_with_dollar(unique)
 
+    def _parse_sort_field(group_dict, sort):
+        field = sort[0][0]
+        sort_order = sort[0][1]
```
gctucker, line 606: What if there are multiple sort parameters?

gctucker: A slightly better way to write this, since sort parameters are 2-tuple, is:
```
field, sort_order = sort[0]
```

app/utils/db.py
```diff
@@ -601,6 +601,14 @@ def _parse_aggregate():
...
+        if sort_order > 0:
+            group_dict['$group'][field] = {'$min': '$'+field}
+        elif sort[0][1] < 0:
```
gctucker, line 609: If `sort_order` can't be 0, then `else` is enough here.

Otherwise, use `sort_order` rather than `sort[0][1]` again.

app/utils/db.py
alexandrasp: a for loop to handle situations when there is more than one field.

app/utils/db.py
alexandrasp: according to documentation just 1 and -1! Updated. :)

gctucker:
Thanks - so I think the code looks good now in principle, let's see if everything works as expected in practice on staging.

PR #247, opened 2020-07-22
T22499 remove boot support

gctucker:
Remove all things related to boot testing from the backend code base.

This won't delete boot results from the Mongo DB, just simplify the backend by removing parts that are effectively dead code now.


gctucker:
Tested OK on staging with this revision: https://staging.kernelci.org/test/job/kernelci/branch/staging.kernelci.org/kernel/staging-20200724.2/

PR #248, opened 2020-07-22
kcidb: use only the SHA1 for revision_id

khilman:
For compatability with kcidb v3 schema, remove the extra namespace
from the revision ID and only use the commit SHA1.


spbnick:
LGTM, thank you, Kevin!

PR #251, opened 2020-08-25
Fix log line translation key error

gctucker:
Fix logic when translating log line numbers, in the case where the log
line number is not in the translation map.

Fixes: db7ab5dae624 ("Clean LAVA test logs")


gctucker:
Tested OK on staging, log lines are still there: https://staging.kernelci.org/test/case/id/5f47df4a766680bbcc448fd5/

And now, kselftest jobs get imported correctly into the database.

PR #260, opened 2020-10-14
db: mongo: timezone-aware connection

khilman:
Connect to mongodb in timezone aware mode.

This enables all timestamps to include timezone information, which is
now required by kcidb schema.

### DIFF
--- a/app/utils/db.py
+++ b/app/utils/db.py
@@ -54,7 +54,8 @@ def get_db_client(db_options):
         db_pool = db_options_get("mongodb_pool", 100)
 
         CLIENT = pymongo.MongoClient(
-            host=db_host, maxPoolSize=db_pool, port=db_port, w="majority")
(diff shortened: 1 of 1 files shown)


khilman:
@gctucker this change, combined with kcidb v8 will fix the timestamp issues which are causing kcidb-submit to fail.

Cc: @spbnick

gctucker:
Thanks, let's see if this solves the kcidb submission issue on staging and look for any regression anywhere else.  We can leave it on staging until next week since only staging is sending data to kcidb at the moment.

spbnick:
Thank you, Kevin! FWIW, this looks good to me.
I can also confirm that both your playground data and the recent data kernelci-backend sends validates correctly now. The recent data also includes all three types of objects: revisions, builds, and tests, and they're all valid. So I would say, **ship it** :grin:!

spbnick (approved):

gctucker:
Great, thanks for confirming.

PR #261, opened 2020-10-22
T24390 Add missing indexes for SoC summary

gctucker:
Add Mongo DB indexes used by queryes on the SoC views.


gctucker:
Tested OK on staging:
https://staging.kernelci.org/soc/

This view now loads in under 1s, it used to take over 10s and sometimes more if a pile of queries were being queued.

PR #264, opened 2020-11-04
Drop the "Not tracking" noisy log messages

gctucker:
The "Not tracking" messages are not too helpful and can be very noisy
with test errors that never get fixed.  Drop them to clen up the
Celery logs.


gctucker:
Tested OK on staging.

PR #265, opened 2020-11-11
Allow UNKNOWN test status as it's reported by LAVA

gctucker:
Under some particular circumstances, LAVA test cases can have the
"UNKNOWN" status.  This also happens to be defined as a status string
in the backend models, so enable it.  Otherwise, an exception is
raised and the results are lost.

### DIFF
--- a/app/models/__init__.py
+++ b/app/models/__init__.py
@@ -332,7 +332,8 @@ VALID_TEST_CASE_STATUS = [
     ERROR_STATUS,
     FAIL_STATUS,
     PASS_STATUS,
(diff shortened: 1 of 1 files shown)


gctucker:
Tested OK on staging, before:
https://staging.kernelci.org/test/plan/id/5fa76da04bffce2ac0a5d245/
after:
https://staging.kernelci.org/test/plan/id/5fad2c012d79cf8497b9c274/

eballetbo:
LGTM. The results look good now.

PR #274, opened 2021-01-11
T25397 handler rework wip

mgalka:
- introduce LavaCallback class to store callback data
- transform data processing into more OOP way
- extract exception raising methods into smaller groups and refactor try
blocks

### DIFF
--- a/app/utils/callback/lava.py
+++ b/app/utils/callback/lava.py
@@ -36,6 +36,8 @@ import yaml
 import json
 import urllib2
 from collections import OrderedDict
+from urlparse import urlparse
+
 
 import utils
 import utils.kci_test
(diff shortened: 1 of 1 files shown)


gctucker:
@mgalka I've rebased the `lava-handler-rework` branch so it now has the GitHub Actions enabled, please rebase this PR so it uses it too.

gctucker (commented):
Please move the 2nd commit to a follow-up PR so the first one can be merged on its own, and drop "WIP:" from the commit summary.

app/models/__init__.py
```diff
@@ -157,6 +157,7 @@
 QEMU_KEY = "qemu"
 RESULT_KEY = "result"
 RETRIES_KEY = "retries"
+ROOTFS_KEY = "rootfs"
```
gctucker: Please factor out the changes with the "rootfs" key, as doing that will require other parts of the code to be updated (emails make use of it).

app/utils/callback/lava.py
```diff
@@ -495,9 +421,7 @@ def __init__(self, job_data, definition_meta, lab_name,
                                         Loader=yaml.CLoader)
             self.log = self._prepare_log(job_data["log"])
         except yaml.YAMLError:
-            ret_code = 401
-            msg = "Invalid test data from LAVA callback"
-            utils.errors.add_error(self._errors, ret_code, msg)
+            raise CallbackParseError
```
gctucker: Might be worth adding an error message to make it clear it's a YAML syntax error.

gctucker (commented):
The code about creating a directory seems to be a recurring pattern, maybe it could be factored out into a utility function.

(these comments will now apply to the new PR with the 2nd commit...)

app/models/__init__.py
gctucker: This now applies to #275

app/utils/callback/lava.py
gctucker: This now applies to #275

gctucker:
Looks fine, still needs to be tested for detecting any regressions as this code should be functionally identical to the main version.

PR #275, opened 2021-01-21
T25397 test groups rework

mgalka:
- Add LavaResult class to store groups and test cases tree
- Remove artifacts storing from LavaCallback class
- Move exception handling to add_tests() function and add
    handle_exceptions() function that takes care about logging and raising
    BackendException

### DIFF
--- a/app/models/__init__.py
+++ b/app/models/__init__.py
@@ -39,6 +39,7 @@ ACCEPTED_KEYS = "accepted"
 ADDRESS_KEY = "address"
 AGGREGATE_KEY = "aggregate"
 ARCHITECTURE_KEY = "arch"
+ATTACHMENTS_KEY = 'attachments'
 BASELINE_KEY = "baseline"
 BOARD_INSTANCE_KEY = "board_instance"
 BOARD_KEY = "board"
--- a/app/utils/__init__.py
+++ b/app/utils/__init__.py
@@ -207,3 +207,12 @@ def get_defconfig_full(
             defconfig_full = defconfig_full_k
 
     return defconfig_full
+
+
+def make_path(dir_path):
+    if not os.path.isdir(dir_path):
+        try:
+            os.makedirs(dir_path)
+        except OSError as e:
+            if e.errno != errno.EEXIST:
+                raise e
--- a/app/utils/callback/lava.py
+++ b/app/utils/callback/lava.py
@@ -61,24 +61,6 @@ LAVA_JOB_RESULT = {
     CANCELING: "UNKNOWN",
 }
 
-TEST_CASE_MAP = {
-    models.NAME_KEY: "name",
-    models.STATUS_KEY: "result",
-}
-
-TEST_CASE_GROUP_KEYS = [
-    models.ARCHITECTURE_KEY,
-    models.BUILD_ENVIRONMENT_KEY,
-    models.DEFCONFIG_FULL_KEY,
-    models.DEVICE_TYPE_KEY,
-    models.GIT_BRANCH_KEY,
-    models.GIT_COMMIT_KEY,
-    models.JOB_KEY,
-    models.KERNEL_KEY,
-    models.LAB_NAME_KEY,
-    models.MACH_KEY,
-]
-
 TEST_CASE_NAME_EXTRA = {
     "http-download": ["label"],
     "git-repo-action": ["commit", "path"],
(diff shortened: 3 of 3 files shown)


app/models/__init__.py
```diff
@@ -157,6 +157,7 @@
 QEMU_KEY = "qemu"
 RESULT_KEY = "result"
 RETRIES_KEY = "retries"
+ROOTFS_KEY = "rootfs"
```
gctucker, line 160: Please factor out the changes with the "rootfs" key, as doing that will require other parts of the code to be updated (emails make use of it).

app/utils/callback/lava.py
```diff
@@ -495,9 +419,7 @@ def __init__(self, job_data, definition_meta, lab_name,
                                         Loader=yaml.CLoader)
             self.log = self._prepare_log(job_data["log"])
         except yaml.YAMLError:
-            ret_code = 401
-            msg = "Invalid test data from LAVA callback"
-            utils.errors.add_error(self._errors, ret_code, msg)
+            raise CallbackParseError
```
gctucker, line 422: Might be worth adding an error message to make it clear it's a YAML syntax error - or let it raise a `YAMLError` if it's always going to be due to this.

app/utils/callback/lava.py
```diff
@@ -638,6 +488,183 @@ def store_rootfs_info(self, file_name):
...
+    job_data["lab_name"] = metadata.get("lab_name")
+    job_data["version"] = metadata.get("version")
+    job_data["boot_log_html"] = metadata.get("boot_log_html")
+
+    if not os.path.isdir(dir_path):
+        try:
+            os.makedirs(dir_path)
+        except OSError as e:
+            if e.errno != errno.EEXIST:
+                raise e
```
gctucker, line 526: Seems worth factoring out into a utility function.

app/utils/callback/lava.py
mgalka: Changed to plain yaml.YamlError

app/utils/callback/lava.py
mgalka: extracted to `utils.make_path()` function

app/models/__init__.py
mgalka: reverted the "rootfs" changes

app/utils/__init__.py
```diff
@@ -207,3 +207,12 @@ def get_defconfig_full(
...
     return defconfig_full
+
+
+def make_path(dir_path):
+    if not os.path.isdir(dir_path):
+        try:
+            os.makedirs(dir_path)
+        except OSError as e:
+            if e.errno != errno.EEXIST:
+                raise e
```
gctucker, line 218: Missing line return.

gctucker (commented):
LGTM except one pycodestyle check to fix.

app/models/__init__.py
```diff
@@ -39,6 +39,7 @@
 ADDRESS_KEY = "address"
 AGGREGATE_KEY = "aggregate"
 ARCHITECTURE_KEY = "arch"
+ATTACHMENTS_KEY = 'attachments'
```
gctucker, line 42: BTW I think we never hit this issue before because there aren't any KernelCI tests using "attachments" at the moment.

app/utils/__init__.py
gctucker: @mgalka It's still there - not a big thing, but git will keep complaining about it and it's good practice to avoid it.

app/models/__init__.py
mgalka: What is more, I remember fixing this issue some time ago, so it's both unused and a regression.

app/utils/callback/lava.py
```diff
@@ -394,6 +364,34 @@ def _get_lava_bootloader_meta(self, bl_meta):
...
         return meta
 
+    def _get_directory_path(self):
+        """Create the dir_path from LAVA metadata
+
+        Update the metadata with the storage path of the artifacts.
+        If possible, use the file_server_resource from the metadata.
+
+        :param meta: The boot meta-data.
+        :type meta: dictionary
```
gctucker, line 362: This method doesn't have any arguments.

In fact I think it's probably OK to not have docstrings for private methods.

gctucker (approved):
Just a minor thing about docstrings, but I think they should all get reviewed once this refactoring work has been completed.

PR #276, opened 2021-01-22
T25891 Change status to SKIP when pushing UNKOWN status to kcidb

mgalka:
kcidb doesn't accept 'UNKNOWN' status of the test as a valid value. This leads to an error and data not being uploaded to kcidb
The closes we can get is changing the status to 'SKIP'.

### DIFF
--- a/app/utils/kcidb.py
+++ b/app/utils/kcidb.py
@@ -179,7 +179,19 @@ def push_build(build_id, first, kcidb_options, db_options={}, db=None):
     _submit(kcidb_data, kcidb_options)
 
 
+def _get_test_misc(group, test):
+    misc = {
+        'plan': group[models.NAME_KEY],
+        'plan_variant': group[models.PLAN_VARIANT_KEY],
+        'kernelci_status': test['status']
+    }
+    return misc
+
+
 def push_tests(group_id, kcidb_options, db_options={}, db=None):
+    status_translate = {
+        'UNKNOWN': 'SKIP'
+    }
     if db is None:
         db = utils.db.get_db_connection(db_options)
     collection = db[models.TEST_GROUP_COLLECTION]
@@ -201,10 +213,6 @@ def push_tests(group_id, kcidb_options, db_options={}, db=None):
         group[models.DEVICE_TYPE_KEY],
         group[models.LAB_NAME_KEY]
     )
-    misc = {
-        'plan': group[models.NAME_KEY],
-        'plan_variant': group[models.PLAN_VARIANT_KEY],
-    }
     env_misc = {
         'device': group[models.DEVICE_TYPE_KEY],
         'lab': group[models.LAB_NAME_KEY],
@@ -245,11 +253,11 @@ def push_tests(group_id, kcidb_options, db_options={}, db=None):
                 },
                 'path': test['path'],
                 'description': test_description,
-                'status': test['status'],
+                'status': status_translate.get(test['status'], test['status']),
                 'waived': False,
                 'start_time': test['start_time'],
                 'output_files': output_files,
-                'misc': misc,
+                'misc': _get_test_misc(group, test)
             }
             for test in test_cases
         ],


gctucker:
Seems fair enough.

@mgalka Please check in the logs if the validation is now accepting all the data being sent.

app/utils/kcidb.py
```diff
@@ -245,11 +248,15 @@ def push_tests(group_id, kcidb_options, db_options={}, db=None):
                 },
                 'path': test['path'],
                 'description': test_description,
-                'status': test['status'],
+                'status': status_translate.get(test['status'], test['status']),
                 'waived': False,
                 'start_time': test['start_time'],
                 'output_files': output_files,
-                'misc': misc,
+                'misc': misc.copy().update(
```
gctucker, line 255: No I think the idea is to keep the original status value in `misc` rather than the translated one which is already in `status`.  So the `status_translate` dictionary shouldn't need to be used here.  Instead, the `kernelci-status` value can be put directly when initialising the `misc` directory with the value of `test['status']`.

gctucker (changes requested):

app/utils/kcidb.py
mgalka: You're right I accidentally pushed the wrong version. My bad.
The question that still remains is if we want to have the `kernelci-status` added to `misc` only when the status was translated to something else or always. In case the latter is the correct solution, that indeed we can populate the value while creating the `misc` directory.

gctucker:
Saw this in the backend logs:
```
Jan 25 16:59:59 kernelci-staging kernelci-celery[18385]: jsonschema.exceptions.ValidationError: None is not of type 'object'
Jan 25 16:59:59 kernelci-staging kernelci-celery[18385]: Failed validating 'type' in schema['properties']['tests']['items']['properties']['misc']:
Jan 25 16:59:59 kernelci-staging kernelci-celery[18385]:     {'description': 'Miscellaneous extra data about the test run',
Jan 25 16:59:59 kernelci-staging kernelci-celery[18385]:      'type': 'object'}
```

gctucker:
I believe the error above was with the version of this PR before the last push, so we'll see later today if the errors go away.

mgalka (commented):
> I believe the error above was with the version of this PR before the last push, so we'll see later today if the errors go away.

There was a problem when `misc` field in the test case was updated. I implemented a fix for that and pushed. Let's see how it goes with the next staging run.

app/utils/kcidb.py
```diff
@@ -179,7 +179,19 @@ def push_build(build_id, first, kcidb_options, db_options={}, db=None):
     _submit(kcidb_data, kcidb_options)
 
 
+def _get_misc(group, test):
```
gctucker, line 182: `misc` can also be found in builds and revisions, so maybe this should be renamed to `_get_test_misc()` to avoid ambibuity.

gctucker (commented):
OK I think this is fine, let's wait for some data to be sent to confirm it's all valid and then drop the debug patch.

gctucker:
Still seeing this exception:
```
Jan 26 07:31:26 kernelci-staging kernelci-celery[23713]:   File "/usr/lib/python3/dist-packages/jsonschema/validators.py", line 130, in validate
Jan 26 07:31:26 kernelci-staging kernelci-celery[23713]:     raise error
Jan 26 07:31:26 kernelci-staging kernelci-celery[23713]: jsonschema.exceptions.ValidationError: None is not of type 'object'
Jan 26 07:31:26 kernelci-staging kernelci-celery[23713]: Failed validating 'type' in schema['properties']['tests']['items']['properties']['misc']:
Jan 26 07:31:26 kernelci-staging kernelci-celery[23713]:     {'description': 'Miscellaneous extra data about the test run',
Jan 26 07:31:26 kernelci-staging kernelci-celery[23713]:      'type': 'object'}
Jan 26 07:31:26 kernelci-staging kernelci-celery[23713]: On instance['tests'][0]['misc']:
Jan 26 07:31:26 kernelci-staging kernelci-celery[23713]:     None
```
So adding `staging-skip` for now as it looks like this is the reason why there haven't been any test results on staging since last night.

app/utils/kcidb.py
mgalka: cahnged to `_get_test_misc()`

mgalka:
> Still seeing this exception:
> 
> ```
> Jan 26 07:31:26 kernelci-staging kernelci-celery[23713]:   File "/usr/lib/python3/dist-packages/jsonschema/validators.py", line 130, in validate
> Jan 26 07:31:26 kernelci-staging kernelci-celery[23713]:     raise error
> Jan 26 07:31:26 kernelci-staging kernelci-celery[23713]: jsonschema.exceptions.ValidationError: None is not of type 'object'
> Jan 26 07:31:26 kernelci-staging kernelci-celery[23713]: Failed validating 'type' in schema['properties']['tests']['items']['properties']['misc']:
> Jan 26 07:31:26 kernelci-staging kernelci-celery[23713]:     {'description': 'Miscellaneous extra data about the test run',
> Jan 26 07:31:26 kernelci-staging kernelci-celery[23713]:      'type': 'object'}
> Jan 26 07:31:26 kernelci-staging kernelci-celery[23713]: On instance['tests'][0]['misc']:
> Jan 26 07:31:26 kernelci-staging kernelci-celery[23713]:     None
> ```
> 
> So adding `staging-skip` for now as it looks like this is the reason why there haven't been any test results on staging since last night.

I've already push a patch that fixes this issue. I'll test it on my local setup and paste results.

gctucker:
OK there's another issue on staging preventing jobs to be started at all, that also explains where there's no results.  Once that has been fixed, this PR can be enabled again and we should be able to verify the exception is gone.

gctucker (approved):
Tested OK on staging, for example here's a test case with "skipped" status:
https://staging.kernelci.org:3000/d/test/test?orgId=1&var-dataset=kernelci04&var-id=kernelci:staging.kernelci.org:6017e93264b4446734816398

PR #279, opened 2021-02-22
T25397 log fragments mixin

mgalka:
Created LogFragmentsMixin class that which is used to group test fragments handling logic. It's content is used by LavaResult class to  handle log fragments adding to the groups structure. The main reason of  mixin creation was to separate log fragments handling from test results.

### DIFF
--- a/app/utils/callback/lava.py
+++ b/app/utils/callback/lava.py
@@ -119,107 +119,12 @@ def _get_definition_meta(meta, job_meta, meta_data_map):
                            " result.".format(ex))
 
 
-def add_log_fragments(groups, log, end_lines_map, start_log_line):
-    lines_map = _prepare_lines_map(end_lines_map, start_log_line)
-    for path, tc in _test_case_iter(groups):
-        try:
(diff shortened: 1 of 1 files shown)


gctucker:
@mgalka The subject of the commit message reads:

> Rework log fragments adding

Does this mean, rework how log fragments get added? Or is it missing some words?

gctucker (changes requested):
This code doesn't work, I get this error:
```
[   ERROR/MainThread] Task lava-test[68fdf64b-15ec-4459-97ff-1fb26e3e9005] raised unexpected: AttributeError("'LavaCallback' object has no attribute 'meta'",)
```

The `meta` attribute is created by the `_prepare_meta()` method, which in turn calls `_get_directory_path()` which uses `meta`.  So there's no way this can work.  #284 seems to be fixing some of that, but it's conflicting so I can't test it.  Please squash the fix in this PR so it can be actually tested and then merged.

gctucker:
Tested OK with fix from #284, both can be merged now on the topic branch.

PR #280, opened 2021-02-26
Add persistent kcidb connection per celery worker

mgalka:
During celery worker startup there are KcidbSubmit objects created for each worker.
When a celery task that needs to push data to KCIDB is run a kcidb-submit process is spawned by the KcidbSubmit object.
These processes live until owning celery worker is terminated.
Tasks that communicate with kcidb stream data to the appropriate kcidb-submit process stdin through a pipe.

### DIFF
--- a/app/taskqueue/tasks/kcidb.py
+++ b/app/taskqueue/tasks/kcidb.py
@@ -17,6 +17,7 @@
 
 """All kcidb related tasks."""
 
+import os
 import taskqueue.celery as taskc
 import utils
 import utils.kcidb
@@ -26,10 +27,12 @@ import utils.kcidb
 def push_build(args):
     build_id, job_id, first = args
     kcidb_options = taskc.app.conf.get("kcidb_options")
-
+    pid = os.getpid()
+    kcidb_submit = taskc.app.kcidb_pool.get(pid)
     if kcidb_options:
         try:
             utils.kcidb.push_build(build_id, first, kcidb_options,
+                                   kcidb_submit,
                                    taskc.app.conf.db_options)
         except Exception, e:
             utils.LOG.exception(e)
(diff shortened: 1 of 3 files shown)


app/taskqueue/tasks/kcidb.py
```diff
@@ -40,10 +43,15 @@ def push_build(args):
 @taskc.app.task(name="kcidb-tests")
 def push_tests(group_id):
     kcidb_options = taskc.app.conf.get("kcidb_options")
+    pid = os.getpid()
+    submit_process = taskc.app.kcidb_pool.get(pid)
+    print(submit_process)
```
gctucker, line 48: Please drop `print` statements once this has been tested on staging, or convert to actual log messages.

app/taskqueue/tasks/kcidb.py
```diff
@@ -40,10 +43,15 @@ def push_build(args):
 @taskc.app.task(name="kcidb-tests")
 def push_tests(group_id):
     kcidb_options = taskc.app.conf.get("kcidb_options")
+    pid = os.getpid()
+    submit_process = taskc.app.kcidb_pool.get(pid)
```
gctucker, line 47: Either use `[pid]` if it's guaranteed to be there, or check if `submit_process` is `None` before using it.

I guess the expectation is that `[pid]` should always work, as this shouldn't get called unless the hooks have been invoked when starting a worker process.

gctucker (commented):
Nice, just a few small things.

app/taskqueue/tasks/kcidb.py
mgalka: I changed it to `[pid]`, it was a leftover from development phase of this fix.

mgalka:
tested locally

PR #281, opened 2021-03-15
T25311 bmeta update

gctucker:
Update the handling of the build meta-data to match the new schema from kernelci-core.


gctucker:
Tested OK on staging, with a single build using the bmeta-rework branch and https://github.com/kernelci/kernelci-core/pull/633:
https://staging.kernelci.org/build/id/606643c6fec669b8adeada9c/

app/utils/__init__.py
```diff
@@ -32,10 +32,10 @@
 LOG = utils.log.get_log()
 
 # Build log file names.
-BUILD_LOG_FILE = "build.log"
-BUILD_ERRORS_FILE = "build-errors.log"
-BUILD_WARNINGS_FILE = "build-warnings.log"
-BUILD_MISMATCHES_FILE = "build-mismatches.log"
+BUILD_LOGS_DIR = "logs"
+BUILD_ERRORS_FILE = "logs/build-errors.log"
```
mgalka, line 36: Wouldn't it be better to change hardcoded paths to something like 
`BUILD_ERRORS_FILE = os.path.join(BUILD_LOGS_DIR, "build-errors.log")` ?

mgalka (changes requested):

app/utils/__init__.py
gctucker: Good point, I shall fix that.

mgalka (approved):
Looks good, thanks!

gctucker:
Fixed issue with getting the modules and dtbs artifacts.  With a kernel build that failed to build the modules, the "modules" URL was using the log file instead:
https://staging.kernelci.org/build/id/608052c1d0f799ea88c7d199/
Now it's fixed, there is no modules as that step failed and the link is not shown:
https://staging.kernelci.org/build/id/608296e2e2716e777ece3084/

PR #288, opened 2021-06-26
Replace invalid characters in test names

mgalka:
There are several characters that may appear in test case or test groups
names, which can be problematic for kcidb or for further processing in
general. This commit takes care of replacing all invalid characters with
underscores.


gctucker (approved):
Tested OK on staging:
https://staging.kernelci.org/test/plan/id/60d915e2f0b47b7dc31eb0b9/

Test cases such as `kselftest-lib.lib.printf.sh` have now been replaced with `kselftest-lib.lib_printf_sh`.

PR #298, opened 2022-12-20
bisect: add a link to the dashboard in bisect email

alexandrasp:
The idea of this PR is to provide in the bisect report a link to the dashboard (https://linux.kernelci.org/test/case/id/6397b076556bd83b952abd03/)  aiming to facilitate the work that Shreeya is doing reporting regressions to the kernel community. 

Signed-off-by: Alexandra Pereira <alexandra.pereira@collabora.com>


alexandrasp:
@mgalka I didn't get a proper way to test this easily locally; I would be happy if you have any suggestions to do this locally. I tried to build a separate instance locally with ansible, but for some reason, I could not understand why Nginx is running the app but does not have permission, I checked all possible logs and permissions then I gave up..because it was taking too much time. I believe this simple change will not break anything and the link will be broken in the worst case.

nuclearcat:

Issue #303, opened 2024-03-15
Add a warning to the email reports telling users that legacy system will shutdown


padovan:
We starting to plan the shutdown of the legacy system (no exact dates yet), so we want to have a warning in the current legacy email report telling the users and asking them to share with us their usecases if they are still using the reports

A message more or less like this:

> This legacy KernelCI providing this report will shutdown sometime soon in favor of our new KernelCI system. Not all tests are being migrated. If you are still using it and want us to prioritize your usecase in the new system, please let us know at kernelci@lists.linux.dev

padovan:
Completed. And we already got one message in the mailing list.

PR #304, opened 2024-04-29
templates: Add shutdown warning

nuclearcat:
Add shutdown warning and suggest users to contact us if they are still using them.

Related-to: https://github.com/kernelci/kernelci-backend/issues/303


padovan:
It is happening! First steps into shutdown.  :)

padovan:
It doesn't seem the warning is appearing in the reports:

* https://groups.io/g/kernelci-results/message/53977
* https://groups.io/g/kernelci-results/message/53975

nuclearcat:
I will verify deploy scripts, it seems they are not functional for more than 2 years for frontend and backend and such things was done manually

